PR #115, opened 2024-06-28
Tweaks

nichochar:
* Fix some border-radius issues on home page
* Fix some minor text color issues
* Added dashed border to upload div. Not perfect but better than solid in my opinion
* Implemented the new "Add cell" with a menu on the left". Tbh it doesn't improve the distancing problem much, which is mostly due to the enormous padding on top of markdown cells. I removed a useless one at the bottom of MD cells (I think it was a mistake), but we still have the problem at the top to be looked at by design.
* fixed some sonner designs (no longer bordering on the div + added X button)

CleanShot 2024-06-28 at 13 28 13

### DIFF
--- a/packages/web/src/routes/session.tsx
+++ b/packages/web/src/routes/session.tsx
@@ -173,8 +173,8 @@ function Session(props: { session: SessionType; channel: SessionChannel }) {
                   <div className="flex-grow border-t"></div>
                   <div className="w-2"></div>
                 </div>
-                <div className="absolute -translate-x-full -top-1/2 flex opacity-0 group-hover:opacity-100 h-9 ">
-                  <div className="flex">
+                <div className="absolute -translate-x-full -top-1/2 flex opacity-0 group-hover:opacity-100 h-10">
+                  <div className="flex items-center">
                     <div className="flex items-center border rounded-sm">
                       <Button
                         variant="secondary"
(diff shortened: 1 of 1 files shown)


packages/web/src/components/srcbook-cards.tsx
```diff
@@ -47,7 +47,7 @@ export function MainCTACard(props: { title: string; description: string; onClick
       className="flex flex-col items-center cursor-pointer border hover:border-foreground transition-colors active:translate-y-0.5 rounded-sm"
       onClick={props.onClick}
     >
-      <div className="w-full grow h-44 bg-border"></div>
+      <div className="w-full grow h-44 bg-border rounded-t-[2px]"></div>
```
benjreinhart, line 50: Why is this 2px, didn't we switch to 3?

packages/web/src/components/srcbook-cards.tsx
```diff
@@ -242,7 +242,7 @@ export function ImportSrcbookCTA(props: { onClick: () => void }) {
   return (
     <CardContainer
       onClick={props.onClick}
-      className="w-full sm:w-[214px] sm:max-w-[214px] bg-muted hover:border-foreground focus-within:border-foreground focus-within:ring-1 focus-within:ring-foreground active:translate-y-0.5"
+      className="w-full sm:w-[214px] sm:max-w-[214px] bg-muted border-dashed hover:border-solid hover:border-foreground focus-within:border-foreground focus-within:ring-1 focus-within:ring-foreground active:translate-y-0.5"
```
benjreinhart, line 245: We can't use `dashed` because it is inconsistent with the dashes we use from the designs (Justin's svg). This looks bad with two different dashes

nichochar: I know it's inconsistent but I think it looks better than not dashed, this was a personal taste call:
1. They both don't look great
2. Dashed at least conveys the right UX, so it wins out

benjreinhart: the solid border looks fine to me and it we use it elsewhere. What is the "right ux" ?

nichochar: Noticed a few times now that we just have different taste for some small things. Not very productive.

Right UX = things with dashed borders imply you can drag things into them

PR #119, opened 2024-07-05
Introduce tsserver for typescript code cells

benjreinhart:
This PR replaces `tsc` with `tsserver` for reporting TypeScript diagnostics. Additionally, it does it on source change, rather than only at the time the code is executed, which makes for a more responsive editor experience.

This is the first step in a larger effort to have a full LSP running in srcbook with type definitions, autocomplete, reporting, etc.

### Prerequisites before merging

- [X] Move all user code files under `src` directory (https://github.com/srcbookdev/srcbook/pull/121)
- [X] Make every typescript project have a tsconfig file inside the project directory (https://github.com/srcbookdev/srcbook/pull/122)
- [X] Ensure the state after deleting is propagated

### TODOs

Lots of TODOs but some immediate follow-ups to call out:

1. cell renaming needs to update tsserver state
2. cell addition needs to update tsserver state
3. package.json adding/removing packages needs to get updated in tsserver
4. We're going to want to debounce source code changes. Right now, we run server logic on every key stroke. We'd want to do this regardless of tsserver, but given we're now doing more on the backend on every change, this is more of a reason to reduce the amount of server state updates.
5. We should stop treating TypeScript diagnostics as cell 'output' and instead treat it as first-class code cell metadata. The rules around diagnostics are not the same as the ones for stdout/stderr. This PR made that clear.

Screenshot 2024-07-08 at 2 00 40 PM


packages/api/server/ws.mts
```diff
@@ -309,13 +315,75 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
...
+
+  for (const cell of session.cells) {
+    if (cell.type === 'code') {
+      tsserver.open({
+        file: Path.join(session.dir, cell.filename),
+        fileContent: cell.source,
+      });
+
+      // Do this once on boot. TODO: Move to other means.
+      setTimeout(async () => {
```
benjreinhart, line 340: This will be rewritten somewhere else, it's just an initial poc.

packages/api/index.mts
```diff
@@ -1,5 +0,0 @@
-import app from './server/http.mjs';
```
nichochar, line 1: I think this will break the build

packages/api/server/ws.mts
```diff
@@ -99,11 +107,8 @@ async function cellExec(payload: CellExecPayloadType) {
       jsExec({ session, cell, secrets });
       break;
     case 'typescript':
-      tscExec({ session, cell, secrets });
       tsxExec({ session, cell, secrets });
       break;
-    default:
```
nichochar, line 105: This didn't have a lot of downsides, and would help discover bugs later.

packages/api/server/ws.mts
```diff
@@ -306,16 +282,109 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
     wss.broadcast(`session:${session.id}`, 'cell:updated', {
       cell: findCell(session, payload.cellId),
     });
+
+    return;
+  }
+
+  if (
+    session.metadata.language === 'typescript' &&
```
nichochar, line 290: This feels redundant with the `tsservers.has(session.id)` predicate

packages/api/server/ws.mts
```diff
@@ -306,16 +282,109 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
...
+    result.cell.type === 'code' &&
+    tsservers.has(session.id)
+  ) {
+    const cell = result.cell;
+    const tsserver = tsservers.get(session.id);
+
+    const file = Path.join(session.dir, cell.filename);
+
+    // To update a file in tsserver, close and reopen it. I assume performance of
+    // this implementation is worse than calculating diffs and using `change` command
```
nichochar, line 300: I've heard that diffs are significantly faster, but if this feels fast enough, seems good enough for now. It also has the benefit of not being stateful and therefore not having TSServer state bugs (I see these in NeoVim here and there and need to kick the LSP)

packages/api/server/ws.mts
```diff
@@ -306,16 +282,109 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
...
+
+  if (!session) {
+    throw new Error(`No session exists for session '${payload.sessionId}'`);
+  }
+
+  if (session.metadata.language !== 'typescript') {
+    throw new Error(`tsserver can only be used with TypeScript Srcbooks.`);
+  }
+
+  if (!tsservers.has(session.id)) {
```
nichochar, line 358: Should it restart if it exists? Do you think we'd ever call this to kick it?

packages/api/server/ws.mts
```diff
@@ -306,16 +282,109 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
...
+    const tsserver = tsservers.create(session.id, { cwd: session.dir });
+
+    // Open all code cells in tsserver
+    for (const cell of session.cells) {
+      if (cell.type === 'code') {
+        tsserver.open({
```
nichochar, line 364: What happens if cell gets deletd?

packages/api/tsserver/tsserver.mts
```diff
@@ -0,0 +1,134 @@
...
+  private _seq: number = 0;
+  private readonly process: ChildProcess;
+  private readonly resolvers: Record<number, (value: any) => void> = {};
+
+  constructor(process: ChildProcess) {
+    this.process = process;
+    this.process.stdout?.on('data', (buffer) => {
+      for (const message of parseTsServerMessages(buffer)) {
+        if (message.type === 'response') {
+          this.handleResponse(message);
```
nichochar, line 15: Can we add a bit of documentation on what responses an d events here from the TSServer's point of view?

packages/api/tsserver/tsserver.mts
```diff
@@ -0,0 +1,134 @@
...
+  private get seq() {
+    return this._seq++;
+  }
+
+  private handleResponse(response: tsserver.protocol.Response) {
+    const resolve = this.resolvers[response.request_seq];
+
+    if (!resolve) {
+      console.warn(
+        `Received a response for command '${response.command}' and request_seq '${response.request_seq}' but no resolver was found. Did you forget to register a resolver?`,
```
nichochar, line 32: Not a problem, but this is funny because you say "You" talking to the dev. Usually a passive form like "was a resolver set?" is used

packages/api/tsserver/tsservers.mts
```diff
@@ -0,0 +1,68 @@
...
+ * This object is responsible for managing multiple tsserver instances.
+ */
+export class TsServers {
+  private servers: Record<string, TsServer> = {};
+
+  get(id: string) {
+    const server = this.servers[id];
+
+    if (!server) {
+      throw new Error(`tsserver for ${id} does not exist.`);
```
nichochar, line 14: I feel like throwing is a bit heavy here, how about returning `null`? I guess we don't really have a policy for this kind of thing, and there's a `has()` method that can be used to check

nichochar (commented):
Looking good. Indeed using the `output` channel feels wrong, makes sense to migrate to its own channel as a follow up.

packages/api/tsserver/tsserver.mts
benjreinhart: The types are defined in protocol.ts in the typescript code base and we define the exact types we expect in this file already. That's good enough for what we need.

packages/api/server/ws.mts
benjreinhart: I don't think implicitly writing code to account for a potential future problem is a good idea. Either do it explicitly when the problem arises or not at all

benjreinhart: Restarting is also inefficient compared to doing nothing if it's already running

packages/api/server/ws.mts
benjreinhart: For large projects with lots of big open files, I imagine that's true. For a couple files with small amounts of code in them, I imagine it's not that different. We give the source directly, so I don't expect tsserver to interact with the disk. If it's not going to disk, it's all in-memory already. The alternative requires calculating diffs, line numbers, offsets, extracting the change, and then passing that to tsserver for it to then do some more computation.

In either case, this won't have end-user impact and this is far simpler without extra dependencies for diffing. Happy to change in the future if this becomes a problem

packages/api/server/ws.mts
benjreinhart: I don't really care either way, but the type system will catch it. So unless we forget to change the typesystem when adding a new language (and how often will that happen?), then I'm not sure how this helps discover bugs faster.

packages/api/tsserver/tsserver.mts
benjreinhart: Meant to enable this only for development so it doesn't appear on users machines. I just changed it though as we don't currently have the process.env set for production in the API.

packages/api/index.mts
benjreinhart: Yes it will... whoops, didn't meant to do this!

PR #120, opened 2024-07-09
Cleanup outdated comments, rename directory, prebuild steps

benjreinhart:
I wanted to rename the directory the examples are in to also use it for other srcbook code (code that is currently in srcbook.mts). While doing so, noticed some outdated code / comments and also noticed we weren't cleaning dirty `dist` dirs and so outdated code was carrying over.


nichochar:
LGTGM

PR #121, opened 2024-07-09
Nest all srcbook code files under src directory

benjreinhart:
We're nesting code files under the `src` directory because:

1. It keeps the srcbook's directory a bit more organized
2. it helps separate user code from other files
3. Makes things like tsconfig `include`s easy

### Compatibility

**This does mean existing srcbook directories _will not work_, at least not in all cases.** We'll want to give folks a heads up here about this backwards-incompatible change (though we are in alpha for this reason). The solution is to export then re-import the srcbook, or just delete if no longer using.

**The .srcmd format we import from / export to _should still work_ because we do not put filepaths in those, only filenames.**

### Unrelated

While working on this PR, I noticed:

* Now that we always use the same id for session / srcbook directory, we should see if we can kill the `dir` field we pass to the client which should always be the same as the `id` and is more confusing than using `id` (especially because on the backend, the `dir` field is the absolute path, not the basename).

### DIFF
--- a/packages/api/srcbook/index.mts
+++ b/packages/api/srcbook/index.mts
@@ -7,21 +7,29 @@ import type {
   PackageJsonCellType,
   SrcbookMetadataType,
 } from '@srcbook/shared';
-import { encode, decode } from './srcmd.mjs';
-import { toFormattedJSON } from './utils.mjs';
 import { randomid } from '@srcbook/shared';
-import { readdir } from './fs-utils.mjs';
-import { SRCBOOKS_DIR } from './constants.mjs';
-import { EXAMPLE_SRCBOOKS } from './srcbook/examples.mjs';
+import { encode, decode } from '../srcmd.mjs';
+import { toFormattedJSON } from '../utils.mjs';
+import { readdir } from '../fs-utils.mjs';
+import { SRCBOOKS_DIR } from '../constants.mjs';
+import { EXAMPLE_SRCBOOKS } from '../srcbook/examples.mjs';
+import { pathToCodeFile, pathToPackageJson, pathToReadme } from './path.mjs';
+
+function writeCellOnlyToDisk(srcbookDir: string, cell: PackageJsonCellType | CodeCellType) {
+  const path =
+    cell.type === 'package.json'
+      ? pathToPackageJson(srcbookDir)
+      : pathToCodeFile(srcbookDir, cell.filename);
+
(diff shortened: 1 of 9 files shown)


packages/api/srcbook/path.mts
```diff
@@ -0,0 +1,22 @@
+import Path from 'node:path';
```
benjreinhart, line 1: While this doesn't add a ton, I want a single source of truth for this. Feels error prone otherwise and should make future refactors of path related changes easier to find/replace than `Path.join`.

packages/api/srcbook/index.mts
```diff
@@ -139,7 +144,8 @@ export async function createSrcbook(title: string, metadata: SrcbookMetadataType
 
 async function createSrcbookDir(basename: string = randomid()) {
   const dirname = Path.join(SRCBOOKS_DIR, basename);
-  await fs.mkdir(dirname, { recursive: true });
+  await fs.mkdir(dirname);
+  await fs.mkdir(Path.join(dirname, 'src'));
```
nichochar, line 148: Could combine both of these into `fs.mkdir(Path.join(dirname, 'src'), { recursive: true });`

packages/api/srcbook/path.mts
nichochar: I like it, especially if we want to map dir IDs later

packages/api/srcmd.mts
```diff
@@ -138,7 +141,11 @@ export async function decodeDir(dir: string): Promise<DecodeResult> {
     // Let's replace all the code cells with the actual file contents for each one
     for (const cell of cells) {
       if (cell.type === 'code' || cell.type === 'package.json') {
-        const filePath = Path.join(dir, cell.filename);
+        const filePath =
+          cell.type === 'package.json'
+            ? pathToPackageJson(dir)
```
nichochar, line 146: Assuming we later add `tsconfig` as another cell type, I think that maybe we'll want to refactor this to something like:
```
pathToFile(cell, dir)
```

which switches on the cell type internally and goes into `src/` or not accordingly.

nichochar (approved):

packages/api/srcbook/index.mts
benjreinhart: yeah I go back and forth on this since I find recursive version 'clever' in the sense it's implicitly doing two things that are both required, so instead I opted for explicit. I don't care about the perf here since this doesn't happen often is equally instant to a human in either case

PR #122, opened 2024-07-09
Generate tsconfig files for each typescript srcbook

benjreinhart:
This will allow each project to have its own tsconfig and eventually allow customization of these values. It is a prerequisite for [tsserver](https://github.com/srcbookdev/srcbook/pull/119).

This is also a backwards incompatible change. We'll group this with the other backwards incompat change in https://github.com/srcbookdev/srcbook/pull/121

### DIFF
--- a/packages/api/exec.mts
+++ b/packages/api/exec.mts
@@ -1,7 +1,5 @@
 import Path from 'node:path';
 import { spawn } from 'node:child_process';
-import { DEFAULT_TSCONFIG_PATH, DEFAULT_TSCONFIG } from './constants.mjs';
-import { tsConfigToArgs } from './utils.mjs';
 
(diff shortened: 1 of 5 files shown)


packages/api/exec.mts
```diff
@@ -84,7 +82,18 @@ export function tsc(options: NodeRequestType) {
   return spawnCall({
     command: Path.join(cwd, 'node_modules', '.bin', 'tsc'),
     cwd,
-    args: [entry, ...tsConfigToArgs(DEFAULT_TSCONFIG)],
+    args: [
```
benjreinhart, line 85: Hardcoding for now. This is going to go away once the tsserver pr is merged anyways.

PR #124, opened 2024-07-11
Add PostHog analytics

nichochar:
This PR introduces posthog analytics for Srcbook. It's an opt-out mechanism, but we don't collect _any_ PII data. Location tracking is disabled, and the only metrics we collect are behavioral and explicit, in the goal of improving the Srcbook application. 

The user can disable tracking in settings.

I added some copy about this in the README. Down the line we might want a more formal document about the data we collect.

I collect on the node side of things (rather than the web side of things). This is a debatable option. It's great for source of truth, but won't help us understand things like clicks / pageViews, etc... It has the advantage of not being able to be blocked by ad-blockers, and won't show up in the network tab.

I do not collect in dev (I guard on `NODE_ENV === production`)

## Testing

This was tested manually, checking that it behaves correctly when enabled and disabled.

## Events

To begin, we track the following events:
 - session create (with TS/JS metadata)
 - generate with AI
 - added openAI API key (generic "changed a setting" event, obviously we don't track the value)
 - delete Srcbook
 - export Srcbook

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ pnpm dlx srcbook
 
 ## Analytics and tracking
 
-In order to improve Srcbook, we collect some behavior analytics. We don't collect anything personal or identifiable, our goals are simply to improve the application (the code is open-source so you don't have to trust us, you can verify!)
+In order to improve Srcbook, we collect some behavioral analytics. We don't collect anything personal or identifiable, our goals are simply to improve the application. The code is open source so you don't have to trust us, you can verify! You can find more information in our [privacy policy](./PRIVACY-POLICY.md).
 
 If you want to disable tracking, you can do so in the settings page of the application.
 
--- a/README.md
+++ b/README.md
@@ -40,6 +40,12 @@ npx srcbook
 pnpm dlx srcbook
 ```
 
+## Analytics and tracking
+
+In order to improve Srcbook, we collect some behavior analytics. We don't collect anything personal or identifiable, our goals are simply to improve the application (the code is open-source so you don't have to trust us, you can verify!)
+
+If you want to disable tracking, you can do so in the settings page of the application.
+
 ## Development
 
 For development instructions, see [CONTRIBUTING.md](./CONTRIBUTING.md).
--- a/packages/api/drizzle/0003_posthog_analytics.sql
+++ b/packages/api/drizzle/0003_posthog_analytics.sql
@@ -1,0 +1,2 @@
+ALTER TABLE `config` ADD `enabled_analytics` integer DEFAULT true NOT NULL;--> statement-breakpoint
+ALTER TABLE `config` ADD `distinct_id` text DEFAULT 's0llcp0p79kcmf7hajobdg3lhc' NOT NULL;
--- a/packages/api/drizzle/0003_posthog_analytics.sql
+++ b/packages/api/drizzle/0003_posthog_analytics.sql
@@ -1,2 +1,2 @@
 ALTER TABLE `config` ADD `enabled_analytics` integer DEFAULT true NOT NULL;--> statement-breakpoint
-ALTER TABLE `config` ADD `distinct_id` text DEFAULT 's0llcp0p79kcmf7hajobdg3lhc' NOT NULL;
+ALTER TABLE `config` ADD `srcbook_installation_id` text DEFAULT '18n70ookj0p2ht8c3aqfu2qjgo' NOT NULL;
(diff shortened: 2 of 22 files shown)


packages/web/src/routes/settings.tsx
```diff
@@ -118,6 +121,24 @@ function Settings() {
             </div>
           </div>
         </div>
+
+        <div>
+          <h2 className="text-xl pb-2">Analytics tracking</h2>
+          <label className="opacity-70 block pb-2">
+            We track behavior analytics to improve the Srcbook product. We track no personal
```
benjreinhart, line 128: > We track behavioral analytics to improve Srcbook. We do not track any personally identifiable information (PII).

packages/api/drizzle/0003_posthog_analytics.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `config` ADD `enabled_analytics` integer DEFAULT true NOT NULL;--> statement-breakpoint
+ALTER TABLE `config` ADD `distinct_id` text DEFAULT 's0llcp0p79kcmf7hajobdg3lhc' NOT NULL;
```
benjreinhart, line 2: Is there a random id generator in sqlite to use here? I know this default should never actually be used but this will be confusing to read in the database and could lead to a future bug where everyone gets the same id.

Nit: I would prefer to name this `distinct_client_id`

README.md
```diff
@@ -40,6 +40,12 @@ npx srcbook
 pnpm dlx srcbook
 ```
 
+## Analytics and tracking
+
+In order to improve Srcbook, we collect some behavior analytics. We don't collect anything personal or identifiable, our goals are simply to improve the application (the code is open-source so you don't have to trust us, you can verify!)
```
benjreinhart, line 45: behavioral*

open source shouldn't have a hyphen here.

packages/api/posthog-client.mts
```diff
@@ -0,0 +1,80 @@
...
+  private isEnabled: boolean = false;
+  private isInitialized: boolean = false;
+  private initializationPromise: Promise<void> | null = null;
+  private eventQueue: QueuedEvent[] = [];
+
+  private constructor() {
+    this.initializationPromise = this.initialize();
+  }
+
+  public static getInstance(): PostHogClient {
```
benjreinhart, line 23: I think we can remove a lot of this code if we just use top-level await to get the config, define a class that takes the config in the constructor, and export an instance of said class?

benjreinhart (commented):
Two points

1. As far as I can tell, this will not work in production. We do not set `NODE_ENV=production` for the API in production (or if we do, I don't know where it happens).
2. Aren't events conventionally defined as variable names instead of descriptions, like `srcbook:started` instead of "user started the srcbook application" ? In all the metrics products i used, there were short event names that acted as keys and you could associate descriptions with them, but that made it less error prone to reuse events in a code base.

packages/api/drizzle/0003_posthog_analytics.sql
nichochar: Yeah, I thought this name was not great. I can change to the one you suggest, it's not really used anywhere so the length isn't a problem.

packages/api/posthog-client.mts
nichochar: done. I initially wanted to avoid that because top level await bit us with initialization before.. but since you had the same inclination, happy to go that direction

srcbook/bin/start
```diff
@@ -1,3 +1,7 @@
 #!/usr/bin/env node
 
-import '../index.mjs';
+process.env.NODE_ENV = 'production';
```
benjreinhart, line 3: Can we use dotenv here, or some other more standard way? We're going to eventually want to set more flags in prod anyways.

srcbook/bin/start
```diff
@@ -1,3 +1,7 @@
...
+
+(async () => {
+  await import('../index.mjs');
```
benjreinhart, line 6: Is using `await` here necessary? And if it is, why do we need to wrap it in a function? We use node with top-level await I thought?

nichochar: I'm not sure about using `await`, but the dynamic import is important, otherwise it gets run before setting the process.env.NODE_ENV.

I don't think there's anything wrong with the current syntax?

srcbook/bin/start
nichochar: Mind if I do this in a follow up? This PR is already blown up a bit.

PR #127, opened 2024-07-11
Treat diagnostics as separate entity from stdout/stderr

benjreinhart:
This PR sends tsserver diagnostics over a separate websocket event as a rich diagnostic object rather than a string message. This paves the way to do things like add squiggly lines under sections with errors, or highlight lines with errors.

I made one other change in this PR which has bugged me for a while: When you clear output, you should only be clearing the output of the current tab, not all tabs, which is confusing and annoying if you didn't mean to do that.


packages/web/src/components/cell-output.tsx
```diff
@@ -66,9 +69,9 @@ export function CellStdio({ cell, show, setShow }: PropsType) {
                     'border-transparent data-[state=active]:border-transparent data-[state=active]:text-tertiary-foreground mb-0',
                 )}
               >
-                {tscOutput.length > 0 ? (
+                {diagnostics.length > 0 ? (
```
benjreinhart, line 72: We call this "problems" but right now we're sending all diagnostics to this tab. That includes warnings and suggestions. So... TODO: rework "Problems" tab to be something like "Diagnostics" which has different categories of diagnostics (error, warning, suggestion)

PR #132, opened 2024-07-14
Use tsserver geterr command & events. reloadProjects for file renames

benjreinhart:
Lots of trial and error but looks like using `geterr` command is better than the `semanticDiagnosticSync` command:

1. It's more convenient since it accepts a list of files rather than one file
2. It sometimes fires events unprompted and so we can listen to those as well as ones we ask it for
3. Of the very little literature i've found on the internet about tsserver, I've seen this referenced

This PR also solves one annoying bug when renaming files

1. Rename a `a.ts` that is imported by `b.ts` to `c.ts`
2. Semantic diagnostics report an error in `b.ts` that `a.ts` doesn't exist
3. Great, all works so far.
4. Rename `c.ts` back to `a.ts`.
5. Semantic diagnostics still report an error in `b.ts` that `a.ts` doesn't exist.
6. This is wrong, `a.ts` does exist.

This PR issues the `reloadProjects` command which gets it back to a working state.

I also added more documentation.

### DIFF
--- a/packages/api/server/ws.mts
+++ b/packages/api/server/ws.mts
@@ -43,7 +43,7 @@ import {
 import tsservers from '../tsservers.mjs';
 import { TsServer } from '../tsserver/tsserver.mjs';
 import WebSocketServer from './ws-client.mjs';
-import { pathToCodeFile } from '../srcbook/path.mjs';
+import { filenameFromPath, pathToCodeFile } from '../srcbook/path.mjs';
 import { normalizeDiagnostic } from '../tsserver/utils.mjs';
 import { removeCodeCellFromDisk } from '../srcbook/index.mjs';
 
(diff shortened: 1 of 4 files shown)


packages/api/server/ws.mts
```diff
@@ -312,12 +312,29 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
     tsserver.close({ file: oldFilePath });
     tsserver.open({ file: newFilePath, fileContent: cellAfterUpdate.source });
 
-    // Send all diagnostics so that other cells that import this updated cell
-    // are informed about any updates (changes to exports, file renames).
-    //
-    // TODO: Can we be smarter and only do this for cells that import
-    // this cell, rather than all cells?
-    sendAllTypeScriptDiagnostics(tsserver, session);
+    // TODO: Given the amount of differences here and elsewhere when renaming cells,
```
nichochar, line 315: +1, I was thinking the same thing

packages/api/server/ws.mts
```diff
@@ -312,12 +312,29 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
...
+      // 4. Rename `c.ts` back to `a.ts`.
+      // 5. Semantic diagnostics still report an error in `b.ts` that `a.ts` doesn't exist.
+      // 6. This is wrong, `a.ts` does exist.
+      //
+      // If we reload the project, this issue resolves itself.
+      //
+      // NOTE: reloading the project sends diagnostic events without calling `geterr`.
+      // However, it seems to take a while for the diagnostics to be sent, so we still
+      // request it below.
+      //
```
nichochar, line 333: nanonit: extra line here

packages/api/tsserver/tsserver.mts
```diff
@@ -1,14 +1,44 @@
...
+ * tsserver is not documented. Here is a brief overview.
+ *
+ * tsserver is a process which listens for messages over stdin and
+ * sends messages over stdout. tsserver has three types of messages:
+ *
+ * 1. Request: A request from the client to the server.
+ * 2. Response: A response from the server to a specific client request.
+ * 3. Event: An event from the server to the client.
+ *
+ * Request and responses are identified a unique number called `seq`. `seq` is incremented
```
nichochar, line 20: identified by* a unique

nichochar (commented):
This is really interesting, thanks for the thorough documentation. Sounds like a blog post on working with tsserver would be really high value ?

PR #134, opened 2024-07-15
Rework cell creation flow

benjreinhart:
This PR:

1. Moves cell creation (insert new cell) to a websocket event like all other cell operations
2. Does not flush the whole session to disk on creation. The only disk writes are: 1) the new cell's file and 2) the updated readme
3. Updates tsserver when a new cell is created, making tsserver aware of the new file. This fixes a scenario where the user has an import statement with a non-existent import path and they then add the file with that path (previously it would still show an error the import doesn't exist when it does. This is now fixed)
4. Updates the insert cell UI component design to the new design (see below)

Screenshot 2024-07-14 at 8 16 05 PM

### DIFF
--- a/packages/web/src/routes/session.tsx
+++ b/packages/web/src/routes/session.tsx
@@ -6,8 +6,13 @@ import {
   CellUpdatedPayloadType,
   CellUpdateAttrsType,
   TsServerCellDiagnosticsPayloadType,
+  CodeLanguageType,
+  MarkdownCellType,
+  CodeCellType,
+  TitleCellType,
+  PackageJsonCellType,
 } from '@srcbook/shared';
-import { loadSession, createCell } from '@/lib/server';
+import { loadSession } from '@/lib/server';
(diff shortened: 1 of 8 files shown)


packages/web/src/routes/session.tsx
```diff
@@ -142,129 +147,103 @@ function Session(props: { session: SessionType; channel: SessionChannel }) {
...
+  const [titleCell, packageJsonCell, ...remainingCells] = allCells;
+  const cells = remainingCells as (MarkdownCellType | CodeCellType)[];
+
   return (
     <>
       <SessionMenu session={session} />
 
       {/* At the xl breakpoint, the sessionMenu appears inline so we pad left to balance*/}
       <div className="px-[72px] xl:pl-[100px]">
+        <TitleCell cell={titleCell as TitleCellType} onUpdateCell={onUpdateCell} />
```
nichochar, line 165: Much nicer to single them out as entities here

PR #137, opened 2024-07-16
AI cell generation

nichochar:
Introduce the ability to create cells (code, markdown) with AI.

Below is a video of the MVP experience. A few things to note:
* I built it so that it can create either markdown or code. There was little in the design that prevented this, and technically it was natural. 
* We need to figure out how to give more context & data when things go wrong. Our users are developers so they'll immediately want to see logs, that's at least what I want.

https://www.loom.com/share/91e4905a691f424db2791539ddfaf3e8?sid=683873e2-0fae-4f98-8019-063002ff3b47

### DIFF
--- a/packages/api/prompts/cell-generator-javascript.txt
+++ b/packages/api/prompts/cell-generator-javascript.txt
@@ -1,0 +1,128 @@
+===== BEGIN INSTRUCTIONS CONTEXT =====
+
+You are tasked with generating a Srcbook code cell for the user.
+
+A Srcbook is a JavaScript notebook following a markdown-compatible format called .srcmd. It's an interactive and rich way of programming that follows the literate programming idea. 
+
+## Srcbook spec
+
+Structure of a Srcbook:
+0. The language comment: `<!-- srcbook:{"language":"javascript"} -->`
+1. Title cell (heading 1)
+2. Package.json cell, listing deps
+3. N more cells, which are either:
+  a. Markdown cells (GitHub flavored Markdown)
+  b. JavaScript code cells, which have a filename and source content.
+
+The user is already working on an existing Srcbook, and is asking you to create exactly one cell at the given position described below. The cell can be code and markdown. If unspecified lean towards a code cell. 
+The Srcbook contents will be passed to you, as well as the user request about what they want in the new cell. Your job is to write the code cell or the markdown cell for the user, making sure you honor their request while leveraging the context of the rest of the Srcbook.
+Each code cell needs to have a unique filename, as it maps to a file on disk.
+
+Code cells are valid javascript code. They have a unique filename. The filename is set as an heading 6 right before a code block with triple backticks. These backticks denote a code block and specify the language, which is always javascript. Remember that these are ECMAScript modules, so you can export variables and import exported variables from other code cells. For example.
+
+Markdown cells are regular markdown. Just avoid using heading1 and heading6, as those are reserved by the Srcbook spec.
+===== END INSTRUCTIONS CONTEXT ======
+
+===== BEGIN EXAMPLE SRCBOOK =====
+<!-- srcbook:{"language":"javascript"} -->
+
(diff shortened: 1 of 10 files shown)


packages/web/src/components/cells/generate-ai.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+        <div className="flex items-start">
+          <Sparkles size={16} className="m-2.5" />
+
+          <textarea
+            value={prompt}
+            autoFocus
+            onChange={(e) => setPrompt(e.target.value)}
+            placeholder="Write a prompt..."
+            className="flex min-h-[60px] w-full rounded-sm px-3 py-2 text-sm placeholder:text-muted-foreground focus-visible:outline-none pl-0"
```
benjreinhart, line 103: Why not codemirror here so we can get auto-resize? Otherwise, we should make this a bit bigger and add a `resize-none`

packages/api/server/http.mts
```diff
@@ -128,6 +128,25 @@ router.post('/generate', cors(), async (req, res) => {
   }
 });
 
+// Generate a cell using AI from a query string
+router.options('/sessions/:id/generate_cell', cors());
```
benjreinhart, line 132: Why not a websocket event?

packages/api/prompts/cell-generator-javascript.txt
```diff
@@ -0,0 +1,128 @@
+===== BEGIN INSTRUCTIONS CONTEXT =====
+
+You are tasked with generating a Srcbook code cell for the user.
+
+A Srcbook is a JavaScript notebook following a markdown-compatible format called .srcmd. It's an interactive and rich way of programming that follows the literate programming idea.
```
benjreinhart, line 5: >  It's an interactive and rich way of programming that follows the literate programming idea.

These prompts feel too wordy IMO. This ^, for example, doesn't seem necessary.

benjreinhart (approved):
IMO, the prompts look too long and wordy to me. There seems to be a lot of unnecessary text in them (why do we need that lengthy example?).

packages/api/server/http.mts
nichochar: I don't have a super strong reason: I favor http when i need request/response semantics with a pending state. It's also less code.

Do you think this would be better as websocket?

packages/api/prompts/cell-generator-javascript.txt
nichochar: Yes, the prompts can definitely be iterated on. This made more sense in the generate srcbook one...I'll remove that particular sentence

Without evals, it's a bit of a nacked process. I am happy to remove this, but the next step for prompt iteration is to set up some basic evals. For this particular flow, the evals could programmatically test that we get valid srcbook code back based on a set of known user request examples. 


Once we have that, we can start iterating with a bit more confidence.

packages/web/src/components/cells/generate-ai.tsx
nichochar: > Why not codemirror here so we can get auto-resize?

Codemirror felt heavyweight for what is essentially a simple multi-line input

> we should make this a bit bigger and add a resize-none

I like the resize, it's not too visible, but gives the user control should they be pasting a bigger prompt (unlikely, but possible). Why don't you like it?

nichochar:
> IMO, the prompts look too long and wordy to me. There seems to be a lot of unnecessary text in them (why do we need that lengthy example?).

Do you have data that suggests that shorter prompts work better? I am happy to iterate on the prompts and shorten them, but I think that the example is one of the most valuable parts. LLMs do very well when you give them examples (few shot).

Are your concerns on cost? latency? 
I'm meeting with the codestory people to talk about this stuff, but I don't have a lot of intuitions yet. We need to setup evals and experiment. This is a first pass which does quite well on my empirical QA.

PR #140, opened 2024-07-17
Decouple srcmd from node libs and AI

benjreinhart:


packages/api/srcmd/encoding.mts
```diff
@@ -0,0 +1,90 @@
...
+    encodeTitleCell(titleCell),
+    encodePackageJsonCell(packageJsonCell, options),
+    ...cells.map((cell) => {
+      switch (cell.type) {
+        case 'code':
+          return encodeCodeCell(cell, options);
+        case 'markdown':
+          return encodeMarkdownCell(cell);
+        case 'placeholder':
+          return encodePlacebolderCell(cell);
```
benjreinhart, line 35: @nichochar there shouldn't be a coupling here to AI-specific things, so we can instead represent a placeholder cell as a cell and then render whatever the contents of the cell's text is.

benjreinhart: This also allows it to be used for other things in the future if needed.

packages/api/ai/srcbook-generator.mts
```diff
@@ -19,9 +24,17 @@ const makeGenerateCellSystemPrompt = (language: CodeLanguageType) => {
 };
 
 const makeGenerateCellUserPrompt = (session: SessionType, insertIdx: number, query: string) => {
-  const inlineSrcbookWithPlaceholder = encode(session.cells, session.metadata, {
+  // Make sure we copy cells so we don't mutate the session
+  const cellsWithPlaceholder: CellWithPlaceholderType[] = [...session.cells];
+
+  cellsWithPlaceholder.splice(insertIdx, 0, {
```
benjreinhart, line 30: Since `CellWithPlaceholderType` is a superset of the `CellType`, any existing calls or cell representations will succeed just fine, but here, in the one place we need to do something different for AI, we can leverage this superset type and pass in the cell instead of relying on the encoder to know what to do for AI.

nichochar (approved):
LGTM

PR #142, opened 2024-07-17
Use .src.md extension instead of .srcmd

benjreinhart:
We're now fully just a markdown file (which is good, since we were before too but nothing would recognize us as such).

Now editors, UIs, etc will render us nicely.


packages/web/src/components/srcbook-cards.tsx
```diff
@@ -218,10 +218,10 @@ export function ImportSrcbookButton(props: { onClick: () => void }) {
     >
       <div className="flex flex-col h-full items-start justify-between">
         <Upload size={20} />
-        <div className="flex flex-col items-start gap-0.5">
-          <h5 className="font-medium leading-[18px]">Open Srcbook</h5>
+        <div className="flex flex-col items-start gap-1">
+          <h5 className="font-medium leading-none">Open Srcbook</h5>
           <p className="leading-none text-[13px] text-tertiary-foreground">
-            or drag'n drop <span className="font-medium">.srcmd</span> file
+            or drag 'n drop <span className="font-medium">.src.md</span> file
```
benjreinhart, line 224: @nichochar this space is intentional, the "'n" is an abbreviation / slang for "and"

nichochar (commented):
lots of changes, but looks good

PR #144, opened 2024-07-19
Ai iterate code cell

nichochar:
Diff based experience with unified view and accept / reject mechanics

https://github.com/user-attachments/assets/eec6f484-8eb8-439b-85ed-0174f0c375ec


packages/shared/src/schemas/websockets.ts
```diff
@@ -24,6 +24,12 @@ export const CellUpdatePayloadSchema = z.object({
   updates: CellUpdateAttrsSchema,
 });
 
+export const CellAiGeneratePayloadSchema = z.object({
```
benjreinhart, line 27: Nit: Can we either drop the `cell` in the name here or change these to be `AIGenerateCell...`? This makes it clear they belong to the AI functionality rather than the normal cell operations.

packages/shared/src/schemas/websockets.ts
```diff
@@ -50,6 +56,11 @@ export const CellUpdatedPayloadSchema = z.object({
   cell: CellSchema,
 });
 
+export const CellAiGeneratedPayloadSchema = z.object({
+  cellId: z.string(),
+  output: z.string(),
```
benjreinhart, line 61: Nit: can we name this `source` to match the name of this property on `cell`? It is the source we're getting back here, right?

packages/shared/src/schemas/websockets.ts
nichochar: Will switch the ordering. I'd love it to be shorter and remove `cell`, but there is an `AiGenerateSrcbook` that makes it ambiguous otherwise.

Right now the `AiGenerateSrcbook` is an http endpoint, but that could change.

packages/shared/src/schemas/websockets.ts
nichochar: `source` felt a little wrong to me, it's LLM output, not really a code source.

PR #147, opened 2024-07-19
Better error handling and styles in generate srcbook modal

benjreinhart:
* If generation fails, we show an error message. This PR makes it so that error message is dismissed the next time the user clicks "generate" or "try again"
* Add correct styles for API key warnings
* Add correct styles for errors
* Special case handling of invalid API key error (ran into this locally)
* Add "try again" link to error message. I'm a bit torn on this since I expect most cases probably won't succeed when trying again, but added it anyways.

### When user never set api key

Screenshot 2024-07-19 at 3 49 08 PM

### When user set api key but the key is invalid according to openai

Screenshot 2024-07-19 at 3 48 47 PM

### Generic error

Screenshot 2024-07-19 at 3 49 27 PM


nichochar:
Very nice!

PR #150, opened 2024-07-22
Clean up deps + use catalogs from pnpm

nichochar:
This is the supported way to pin package versions across monorepo sub-packages. See [docs](https://pnpm.io/catalogs)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -25,9 +25,6 @@ jobs:
 
       - uses: pnpm/action-setup@v4
         name: Install pnpm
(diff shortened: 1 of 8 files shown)


.github/workflows/run-tests.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       - uses: pnpm/action-setup@v4
         name: Install pnpm
         with:
-          version: 8
```
nichochar, line 29: This is the build fix, used to compete with the version pinned in our package.json. This script knows how to read the package.json if unspecified

PR #151, opened 2024-07-22
Introduce styled tooltips. Refactor to be cross-platform

nichochar:
Introduced and styled tooltips.
Refactored code around keyboard shortcuts to be more reusable + cross platform

CleanShot 2024-07-22 at 16 37 11
CleanShot 2024-07-22 at 16 37 07
CleanShot 2024-07-22 at 16 35 34
CleanShot 2024-07-22 at 16 35 25


benjreinhart:
We're going to show that on every hover? Isn't this super noisy?

nichochar:
It doesn't feel noisy, but let's live with it and adjust the delay time if necessary before the next release.

PR #153, opened 2024-07-23
Dont show secrets unless action is taken by user

nichochar:
This prevents the security vulnerability of accidentally showing secrets when sharing screen / streaming video.

Now it requires an explicit action from the user (see video).

Also added `cmd + enter` shortcut for submitting the form

https://github.com/user-attachments/assets/0428eb95-3265-4416-905b-c4d699428d7c

### DIFF
--- a/packages/web/src/routes/secrets.tsx
+++ b/packages/web/src/routes/secrets.tsx
@@ -1,6 +1,8 @@
 import { useRef, useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
+import { cn } from '@/lib/utils';
 import { getSecrets } from '@/lib/server';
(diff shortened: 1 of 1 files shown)


packages/web/src/routes/secrets.tsx
```diff
@@ -264,13 +282,23 @@ function SecretRow(props: {
 function NewSecretForm(props: {
   setError: (message: string | null, clearAfter?: number | null) => void;
 }) {
+  useHotkeys(
```
benjreinhart, line 285: In all cases I've seen thus far for inputs, i find this pattern to be less obvious (the code is located somewhere other than the `Input`) and heavier weight than:

```jsx
onKeydown={(e) => e.key === 'Enter' && onSubmit()}
```

PR #155, opened 2024-07-24
Make sure ai import is a type, add zod to deps

nichochar:


benjreinhart (commented):
Why are there changes to pnpm / package.json here?

nichochar:
I added zod to the deps explicitely. 

This isn't broken in the published build, but without this, you cannot run `./bin/start` from the `srcbook/srcbook` repo:
CleanShot 2024-07-23 at 18 08 13

PR #157, opened 2024-07-24
Open srcbook in browser on launch

benjreinhart:
I also added the version and package name so we know exactly what's booting.

Screenshot 2024-07-24 at 1 06 10 PM


nichochar (commented):
LGTM, but yeah lets test this with `scratch`

PR #158, opened 2024-07-25
Add feedback form

nichochar:
After trying to use airtable and getting very annoyed at how clunky it is, I figured out we could simply use google sheets with their "app scripts" API. This is free, and works.

### Note
* I realized that posthog has a "surveys" product. It has some nice things (like NPS score or PMF questionnaire), but as far as I could tell, you have to use their UI, so it's styled inconsistently with the rest of the app, and it says "by posthog" in the middle of it. Holding off of that for now, but good to consider if we want to do more advanced surveys later.

CleanShot 2024-07-27 at 16 14 15@2x

### DIFF
--- a/packages/web/src/components/feedback-dialog.tsx
+++ b/packages/web/src/components/feedback-dialog.tsx
@@ -1,7 +1,8 @@
 import { useState } from 'react';
-import { sendFeedback } from '@/lib/utils';
+import { sendFeedback } from '@/lib/server';
 import { toast } from 'sonner';
 import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
 import {
   Dialog,
   DialogContent,
@@ -19,6 +20,8 @@ export default function FeedbackDialog({
   onOpenChange: (open: boolean) => void;
 }) {
   const [feedback, setFeedback] = useState('');
+  const [email, setEmail] = useState('');
+
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
       <DialogContent className="max-w-xl">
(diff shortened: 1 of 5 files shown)


packages/web/src/lib/utils.ts
```diff
@@ -27,3 +27,18 @@ export function getTitleForSession(session: SessionType) {
   const titleCell = session.cells.find((cell: CellType) => cell.type === 'title') as TitleCellType;
   return titleCell?.text;
 }
+
+// POST feedback to Google Sheets
+// Because of CORS, we don't get a response here so this is a fire and forget.
```
benjreinhart, line 32: Why not make this request from the API where CORS is not an issue?

packages/web/src/components/feedback-dialog.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [feedback, setFeedback] = useState('');
+  return (
+    <Dialog open={open} onOpenChange={onOpenChange}>
+      <DialogContent className="max-w-xl">
+        <DialogHeader>
+          <DialogTitle>Share Feedback</DialogTitle>
+          <DialogDescription asChild>
+            <div className="pt-4 flex flex-col gap-2">
+              <p>
+                We're always looking to improve Srcbook, and your feedback is invaluable.
```
benjreinhart, line 30: There shouldn't be a comma after the "and"

whatrocks:
I'd suggest optionally letting the user leave their email so that you can reply to them

packages/web/src/lib/utils.ts
nichochar: Duh, I was clearly tired as I did not think of this. Fixed.

nichochar:
> I'd suggest optionally letting the user leave their email so that you can reply to them

Good suggestion @whatrocks , added

PR #167, opened 2024-08-06
Add purple colors and 'Use AI in Srcbook nudge'

nichochar:
Following designs, but only had light mode in the design file.
I'm therefore following the pattern that we have for `error` and `warning` colors, which are stable in light/dark mode (I believe that this is how the design system works).
CleanShot 2024-08-05 at 18 17 53
CleanShot 2024-08-05 at 18 18 00

### DIFF
--- a/packages/web/tailwind.config.js
+++ b/packages/web/tailwind.config.js
@@ -94,6 +94,11 @@ module.exports = {
           DEFAULT: 'hsl(var(--warning))',
           foreground: 'hsl(var(--warning-foreground))',
         },
+        ai: {
+          DEFAULT: 'hsl(var(--ai))',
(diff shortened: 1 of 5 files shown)


packages/web/tailwind.config.js
```diff
@@ -139,6 +144,11 @@ module.exports = {
           'blue-60': 'hsl(var(--sb-blue-60))',
           'blue-70': 'hsl(var(--sb-blue-70))',
           'blue-80': 'hsl(var(--sb-blue-80))',
+
+          'purple-10': 'hsl(var(--sb-purple-10))',
```
nichochar, line 148: The way I copied these over was super tedious, wondering if I'm missing something, could not find them easily in figma

PR #168, opened 2024-08-07
Support anthropic in addition to openAI

nichochar:
This is step 2/3 for our early providers support (OpenAI, Anthropic, Local models).

I ended up refactoring quite a bit, since it felt needed.

## Notes

 - settings now uses a context
 - we support 2 provider/models. Currently, you cannot pick the model, but nothing is preventing this except lack of design and me wanting to keep things simple to start.
 - added some green to our colors as-per the design. Similarly to purple, we might want to improve the dark mode version of it which is very high contrast as it stands


packages/api/db/schema.mts
```diff
@@ -1,16 +1,27 @@
 import { sqliteTable, text, integer } from 'drizzle-orm/sqlite-core';
 import { randomid } from '@srcbook/shared';
 
+// Opening us up for different types of configuration:
+// supporting different models, local models and openAI url compatible providers
+type AiConfig =
```
nichochar, line 6: Debated putting these in `@srcbook/shared`. I will sleep on it

packages/api/db/schema.mts
```diff
@@ -1,16 +1,27 @@
...
   baseDir: text('base_dir').notNull(),
   defaultLanguage: text('default_language').notNull().default('typescript'),
   openaiKey: text('openai_api_key'),
+  anthropicKey: text('anthropic_api_key'),
   // Default on for behavioral analytics.
   // Allows us to improve Srcbook, we don't collect any PII.
   enabledAnalytics: integer('enabled_analytics', { mode: 'boolean' }).notNull().default(true),
   // Stable ID for posthog
   installId: text('srcbook_installation_id').notNull().default(randomid()),
+  aiConfig: text('ai_config', { mode: 'json' })
```
nichochar, line 21: recommended way to do json by drizzle orm

PR #175, opened 2024-08-15
Support openai-compatible ai providers. Ability to pick AI models

### LINKED ISSUE #161: Support multiple AI providers (and an OSS option)
Today, we support only 1 provider: openAI.

In the settings, users should be able to pick between:

* OpenAI (they need to specify an API key)
* Anthropic Claude (they need to specify an API key)
* Ollama (they need to specify a url)

We need a UX pattern where they can select which of the AI providers they want, which also makes it clear if AI is enabled or not (at least one API key is provided and that provider is selected).

Not sure yet if we want to use an abstraction like Ollama or LMStudio for people picking their own model, vs a "openAI compatible API url".

nichochar:
We want users to be able to use open-source models locally, with tools like [Ollama](https://ollama.ai).

This PR allows to connect to openai-api-compatible models, by specifying the baseURL. Video of experience below:

https://github.com/user-attachments/assets/cd79a0bb-57b1-49c7-a797-8f58a07b115f

## Notes
* Currently, users cannot change anything but the `baseURL` parameter. In particular this means that we don't yet support adding auth-headers to third party hosted services like Groq, fireworks et.al. We absolutely could support them, but I'm waiting to see if we get customer demand.
* I added a "test AI" button. This isn't yet the `aiEnabled` source of truth, but I think that would be a nice improvement later.
* We don't show logs to the user in the UI if testing fails: we point them to the server logs instead.
* I have not yet updated to the new AI accent color. Will do that in a follow up PR when I update all of the AI stuff.
* This contains a migration. The old column is deprecated, but I'm not deleting in case something goes wrong and we want to roll back.
* I tested a couple of small models and they kind of suck. Mistral-nemo did best. Down the line we could fine-tune one and publish it to ollama...

fixes #161

### DIFF
--- a/packages/api/ai/config.mts
+++ b/packages/api/ai/config.mts
@@ -1,0 +1,44 @@
+import { createOpenAI } from '@ai-sdk/openai';
+import { createAnthropic } from '@ai-sdk/anthropic';
+import { getConfig } from '../config.mjs';
+import type { LanguageModel } from 'ai';
+import { AiProvider, getDefaultModel, type AiProviderType } from '@srcbook/shared';
+
+/**
+ * Get the correct client and model configuration.
+ * Throws an error if the given API key is not set in the settings.
+ */
+export async function getModel(): Promise<LanguageModel> {
+  const config = await getConfig();
+  const { aiModel, aiProvider } = config;
+  const model = aiModel || getDefaultModel(aiProvider as AiProviderType);
+  switch (aiProvider as AiProviderType) {
+    case 'openai':
+      if (!config.openaiKey) {
+        throw new Error('OpenAI API key is not set');
+      }
+
+      const openai = createOpenAI({
+        compatibility: 'strict', // strict mode, enabled when using the OpenAI API
+        apiKey: config.openaiKey,
+      });
+
+      return openai(model);
+    case 'anthropic':
+      if (!config.anthropicKey) {
+        throw new Error('Anthropic API key is not set');
+      }
+      const anthropic = createAnthropic({
+        apiKey: config.anthropicKey,
+      });
+
+      return anthropic(model);
+
+    case 'local':
+      const openaiCompatible = createOpenAI({
+        compatibility: 'compatible',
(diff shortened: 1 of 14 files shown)


packages/api/ai/config.mts
```diff
@@ -0,0 +1,44 @@
...
+      const anthropic = createAnthropic({ apiKey: config.anthropicKey });
+      return anthropic(model);
+
+    case 'custom':
+      if (typeof aiBaseUrl !== 'string') {
+        throw new Error('Local AI base URL is not set');
+      }
+      const openaiCompatible = createOpenAI({
+        compatibility: 'compatible',
+        apiKey: 'bogus', // required but unused
```
benjreinhart, line 39: Why not send an API key (even if possibly null)? For local models it might not be needed but it what if you want to use your company's proxy or a service like together? We're saying this is openai-compatible, but that's not really true if we don't include an API key here (e.g., togetherai has openai compat endpoints, but they need an API key).

packages/api/db/schema.mts
```diff
@@ -1,7 +1,8 @@
 import { sqliteTable, text, integer } from 'drizzle-orm/sqlite-core';
 import { randomid } from '@srcbook/shared';
 
-// Opening us up for different types of configuration:
+// Deprecated. Remove me when migration of AI config is stable.
```
benjreinhart, line 4: We should kill this right away

packages/web/src/routes/settings.tsx
```diff
@@ -186,32 +237,95 @@ function Settings() {
...
+      {aiEnabled ? <TestAiButton /> : fragments(aiProvider)}
+    </div>
+  );
+}
+
+const TestAiButton = () => {
+  const [state, setState] = useState<'idle' | 'loading' | 'success' | 'error'>('idle');
+
+  const TIMEOUT = 2500;
+  const check = () => {
```
benjreinhart, line 288: nit: this should be in a useEffect

PR #183, opened 2024-08-16
Add dynamic build vars for web's index.html

benjreinhart:
Locally we run dev server and API on separate hosts. In prod, these are the same and also should be computed at runtime (window.location.*) to account for localhost aliases.

See https://vitejs.dev/guide/env-and-mode#html-env-replacement

(they can also be overridden with an env file with a `.local` suffix).


nichochar:
LGTM

PR #191, opened 2024-08-20
Support disabling analytics through env var

benjreinhart:
Removed settings page analytics.

To disable analytics, run srcbook with the `SRCBOOK_DISABLE_ANALYTICS=true` env var set

### DIFF
--- a/srcbook/src/server.mjs
+++ b/srcbook/src/server.mjs
@@ -64,7 +64,8 @@ server.listen(port, () => {
 });
 
 process.on('SIGINT', async () => {
-  await posthog.shutdown();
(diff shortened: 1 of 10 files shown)


srcbook/src/server.mjs
```diff
@@ -64,7 +64,6 @@ server.listen(port, () => {
 });
 
 process.on('SIGINT', async () => {
-  await posthog.shutdown();
```
nichochar, line 67: Why are you removing this?

benjreinhart: Didn't see the call to the underlying posthog shutdown

PR #204, opened 2024-08-23
Update cards on homepage

benjreinhart:
Cards previously looked like there was an issue with assets loading. This also adds the tags found on hub and srcbook.com.

Screenshot 2024-08-22 at 6 58 41 PM

Screenshot 2024-08-22 at 6 58 46 PM

### DIFF
--- a/packages/web/src/components/logos.tsx
+++ b/packages/web/src/components/logos.tsx
@@ -5,103 +5,36 @@ type PropsType = {
   className?: string;
 };
 
-/*
- * Our logo comes in different widths based on the size.
- */
 export function SrcbookLogo(props: PropsType) {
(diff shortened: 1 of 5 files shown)


packages/web/src/components/logos.tsx
```diff
@@ -5,103 +5,36 @@ type PropsType = {
   className?: string;
 };
 
-/*
- * Our logo comes in different widths based on the size.
- */
 export function SrcbookLogo(props: PropsType) {
```
nichochar, line 8: This is a regression. There are 3 different logos based on the size of the square, otherwise the ratio of the vertices and the nodes becomes off. We need to reintroduce.

PR #206, opened 2024-08-23
Node 18 support

### LINKED ISSUE #189: support nodejs v18
running command as below:

```
srcbook start
```

cli throw error messages:

```sh
~/.nvm/versions/node/v18.17.0/lib/node_modules/srcbook/node_modules/@srcbook/shared/dist/src/utils.js:3
    const bytes = crypto.getRandomValues(new Uint8Array(byteSize));
                  ^

ReferenceError: crypto is not defined
    at randomid (~/.nvm/versions/node/v18.17.0/lib/node_modules/srcbook/node_modules/@srcbook/shared/dist/src/utils.js:3:19)
    at ~/.nvm/versions/node/v18.17.0/lib/node_modules/srcbook/lib/db/schema.mjs:13:66
```

I see your document nodejs version requirements is v20+.   can you support nodejs v18.

versecafe:
# Fixes #189 

Just references crypto with an import instead of assuming in name space

```diff
+ import * as crypto from 'crypto';

  export function randomid(byteSize = 16) {
    const bytes = crypto.getRandomValues(new Uint8Array(byteSize));
    return base32hexnopad.encode(bytes).toLowerCase();
  }
```


nichochar:
Indeed my cursory exploration of node18 compat only brought up this issue. I will do a bit of QA to make sure most of the functionality works correctly, but otherwise LGTM. 

We're not heavily prioritizing earlier versions FYI, since we might want to use some recent features like replacing tsx with vanilla node with [`--experimental-strip-types`](https://github.com/nodejs/loaders/issues/208).

However, if it's easy in the meantime, I'm all for compatibility.

versecafe:
--experimental-strip-types is probably something I'd recommend avoiding until it's more stable unless you allow users on bun to run the TS proper, and find a solution for people on Deno, and it could be removed at any time

packages/shared/src/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { base32hexnopad } from '@scure/base';
 import type { CodeLanguageType } from './types/cells.js';
+import * as crypto from 'crypto';
```
benjreinhart, line 3: This is used on the client in addition to the Node backend. What happens in that environment? Is vite smart enough to remove this from the compiled code, or does it try to do some sort of polyfill (e.g. the way webpack used to with crypto-browserify)?

srcbook/README.md
```diff
@@ -24,7 +24,7 @@ Srcbook runs locally on your machine as a CLI application with a web interface.
 
 ### Requirements
 
-- Node.js v20+
+- Node.js v18+
```
benjreinhart, line 27: If we do support 18, then we should change CI to run against 18 (earliest version we support) and the current node version? I think it's currently 20 and 22

packages/shared/src/utils.ts
versecafe: This should technically work Vite doesn't do a polyfil but ignores the import which *should* then cause it to end up using the global window crypto instance, or polyfills can be added if for some reason it's not but it does seem to be working

benjreinhart: I do not want us to use polyfills for something that exists natively. So assuming this drops the import then I'm stoked to get this working in node 18.

benjreinhart (approved):
I couldn't find anything in a quick scan of vite's docs. My only concern would be if vite were to include a polyfill in our build that we don't need since I don't want to blow up the client side code for no reason (like webpack would)

I'm going to assume the above isn't the case and approve, but if you have any references in the docs you can provide that'd be reassuring

versecafe:
@benjreinhart it will give a warning of 

```
[plugin:vite:resolve] [plugin vite:resolve] Module "crypto" has been externalized for browser compatibility, imported by "/Users/veronica/Projects/srcbook/packages/shared/dist/src/utils.js". See https://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
```

but this just means it'll fail if you try to call the node version of crypto on the browser since no pollyfill is used, but the isBrowser check means that on a browser it will always end up using `globalThis.crypto` which is built in

Issue #207, opened 2024-08-24
Cap title length of srcbooks


nichochar:
### Objective
Limit Srcbook titles (encoded as markdown h1) to a maximum length for design and pragmatic reasons.

### Requirements
1. Set max title length to N characters (proposed: N = ~~50~~ 44)
2. Implement validation at:
   - Import time
   - Client-side update in /srcbook view
   - Server-side update

benjreinhart:
We already have one area that caps this at 44 characters. I believe that was calculated based on the available space in the cards on the home page and also how it looks visually on the Srcbook page.

PR #201, opened 2024-08-22
Overhaul monorepo tooling

versecafe:
## Checklist
- [x] NVM version file
- [X] Turbo repo build tools
- [x] Turbo cache policies
- [x] GitHub Actions CI
- [X] Turbo caching on CI (just a quick change to CI)
- [x] Remove script mess
- [x] Compile packages cleanly (just cleaner not redoing full build system)

PR #212, opened 2024-08-25
fix: Limit title length to 50 characters in cellUpdate function

Aswanth-c:
Fixed https://github.com/srcbookdev/srcbook/issues/207 

I have replaced  `<h1>` to `<textarea>` because the `contentEditable` approach in `<h1>`  did not handle text wrapping efficiently, which leads to an overflow of text  like below 

CleanShot 2024-08-25 at 01 03 28@2x

### DIFF
--- a/packages/api/server/ws.mts
+++ b/packages/api/server/ws.mts
@@ -416,6 +416,7 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
 
 async function cellRename(payload: CellRenamePayloadType) {
   const session = await findSession(payload.sessionId);
+
   if (!session) {
     throw new Error(`No session exists for session '${payload.sessionId}'`);
   }
--- a/packages/api/server/ws.mts
+++ b/packages/api/server/ws.mts
@@ -390,6 +390,13 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
       `No cell exists for session '${payload.sessionId}' and cell '${payload.cellId}'`,
     );
   }
+  if (
+    cellBeforeUpdate.type === 'title' &&
+    'text' in payload.updates &&
+    payload.updates?.text?.length > 50
+  ) {
+    throw new Error('Title cannot be more than 50 characters');
+  }
 
   const result = await updateCell(session, cellBeforeUpdate, payload.updates);
 
@@ -416,7 +423,6 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
 
 async function cellRename(payload: CellRenamePayloadType) {
   const session = await findSession(payload.sessionId);
-
   if (!session) {
     throw new Error(`No session exists for session '${payload.sessionId}'`);
   }
--- a/packages/api/server/ws.mts
+++ b/packages/api/server/ws.mts
@@ -390,13 +390,6 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
       `No cell exists for session '${payload.sessionId}' and cell '${payload.cellId}'`,
     );
   }
-  if (
-    cellBeforeUpdate.type === 'title' &&
-    'text' in payload.updates &&
-    payload.updates?.text?.length > 50
-  ) {
-    throw new Error('Title cannot be more than 50 characters');
-  }
 
   const result = await updateCell(session, cellBeforeUpdate, payload.updates);
 
--- a/packages/web/src/components/import-export-srcbook-modal.tsx
+++ b/packages/web/src/components/import-export-srcbook-modal.tsx
@@ -32,8 +32,8 @@ export function ImportSrcbookModal({
   async function onChange(entry: FsObjectType) {
     setError(null);
 
-    if (entry.basename.length > 50) {
-      setError('File name should be less than 50 characters');
+    if (entry.basename.length > 44) {
+      setError('Srcbook title should be less than 44 characters');
       return;
     }
 
--- a/packages/web/src/components/import-export-srcbook-modal.tsx
+++ b/packages/web/src/components/import-export-srcbook-modal.tsx
@@ -32,6 +32,11 @@ export function ImportSrcbookModal({
   async function onChange(entry: FsObjectType) {
     setError(null);
 
+    if (entry.basename.length > 50) {
+      setError('File name should be less than 50 characters');
+      return;
+    }
+
     const { error: importError, result: importResult } = await importSrcbook({ path: entry.path });
 
     if (importError) {
--- a/packages/web/src/components/ui/heading.tsx
+++ b/packages/web/src/components/ui/heading.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { cn } from '@/lib/utils';
 import { Info } from 'lucide-react';
-import { z } from 'zod';
+import { TitleCellUpdateAttrsSchema } from '@srcbook/shared';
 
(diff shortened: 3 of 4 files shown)


packages/api/server/ws.mts
```diff
@@ -390,6 +390,13 @@ async function cellUpdate(payload: CellUpdatePayloadType) {
       `No cell exists for session '${payload.sessionId}' and cell '${payload.cellId}'`,
     );
   }
+  if (
```
benjreinhart, line 393: Title validation happens in `session.mts` in a function called `updateTitleCell`. In there, we use a Zod schema to handle validation. That schema currently does not cap the length but that seems like the place to add this validation rule.

Currently, the zod schema parsing will throw an exception if there's an issue. So long as the client also implements  validation, then I think it's fine for the exception to crash since it would have to be some sort of bizarre edge case or bug in client code if that were to happen.

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -32,6 +32,11 @@ export function ImportSrcbookModal({
   async function onChange(entry: FsObjectType) {
     setError(null);
 
+    if (entry.basename.length > 50) {
+      setError('File name should be less than 50 characters');
```
benjreinhart, line 36: Can we call this "Srcbook title" instead of file name?

packages/web/src/components/ui/heading.tsx
```diff
@@ -10,27 +11,51 @@ export function EditableH1(props: {
...
-      onKeyDown={(e) => {
-        if (e.key === 'Enter' && ref.current) {
-          ref.current.blur();
-        }
-      }}
-    >
-      {props.text}
-    </h1>
+    <div>
+      <textarea
```
benjreinhart, line 33: An `h1` is a better element for this purpose in terms of semantics. Can you help me understand why that won't work here? What are the problems with using an h1 and are there ways to overcome them?

benjreinhart (commented):
Thank you for this PR!! I have a couple notes:

I know #207 says 50 characters, but we already have [one area of validation in the code base for this that uses 44 characters](https://github.com/srcbookdev/srcbook/blob/main/packages/api/server/http.mts#L63-L64). I prefer 44 for now because that was chosen for a reason previously, though in either case the validation will need to be consistent.

versecafe:
Why not just allow long names and CSS truncate the text to `...` to allow people to easily paste things in without hitting validation issues?

benjreinhart:
@versecafe we could, and we do this in other places. But given that there's no where in the UI where we render a long title in its entirety, there seems to be no point in allowing one in the first place. IMO, not needing to truncate at all is better than truncating

packages/web/src/components/ui/heading.tsx
benjreinhart: I'm not yet convinced we need to change to using a textarea instead of an h1. If it is the case that we are unable to prevent editing a contenteditable element when the length exceeds N, then we should use an h1 until they focus the element for editing at which point we can change to a textarea. Then we should change it back on blur.

I'll take a quick look at this tomorrow.

packages/web/src/components/ui/heading.tsx
```diff
@@ -10,27 +11,51 @@ export function EditableH1(props: {
...
+          }
+        }}
+        onChange={(e) => {
+          handleChange(e.target.value);
+        }}
+        ref={ref}
+        rows={1}
+      />
+      {isMaxHeadingLengthExceeded && (
+        <div className="bg-error text-error-foreground flex items-center rounded-sm border border-transparent px-[10px] py-2 text-sm leading-none font-medium">
```
benjreinhart, line 54: Can you share a screen shot of what this looks like?

benjreinhart (commented):
Would it be easy to reuse the zod schema in all the places where we do validation? If so, can we make that change? That'll make it easier on our future selves when maintaining this.

packages/web/src/components/ui/heading.tsx
Aswanth-c: CleanShot 2024-08-25 at 22 36 01@2x

packages/web/src/components/ui/heading.tsx
```diff
@@ -10,27 +11,51 @@ export function EditableH1(props: {
   className?: string;
   onUpdated: (text: string) => void;
 }) {
-  const ref = React.useRef<HTMLHeadingElement | null>(null);
+  const ref = React.useRef<HTMLTextAreaElement | null>(null);
+  const [heading, setHeading] = React.useState<string>(props.text);
+  const [isMaxHeadingLengthExceeded, setIsMaxHeadingLengthExceeded] =
+    React.useState<boolean>(false);
+  const maxHeadingLength = 50;
```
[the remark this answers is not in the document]
Aswanth-c: Changed!

packages/web/src/components/ui/heading.tsx
```diff
@@ -1,36 +1,72 @@
...
   className?: string;
   onUpdated: (text: string) => void;
 }) {
-  const ref = React.useRef<HTMLHeadingElement | null>(null);
+  const ref = React.useRef<HTMLHeadingElement>(null);
+  const [isMaxHeadingLengthExceeded, setIsMaxHeadingLengthExceeded] =
+    React.useState<boolean>(false);
+
+  const maxHeadingLength = 44;
+  const headingSchema = z.string().max(maxHeadingLength, {
```
benjreinhart, line 20: A schema for the title is already defined in the shared package, so we should be able to import and reuse that here rather than defining a new one.

Aswanth-c: Done!

benjreinhart (approved):
This allows users to type in a long heading but then it disappears on blur with a clipped version of the text. I think the better UX here would be to prevent them from typing in the long text.

I'm going to merge this for now and can address UX improvement in follow up.

benjreinhart:
Thank you for this 🙏🏻

Aswanth-c:
> This allows users to type in a long heading but then it disappears on blur with a clipped version of the text. I think the better UX here would be to prevent them from typing in the long text.
> 
> I'm going to merge this for now and can address UX improvement in follow up.



> This allows users to type in a long heading but then it disappears on blur with a clipped version of the text. I think the better UX here would be to prevent them from typing in the long text.
> 
> I'm going to merge this for now and can address UX improvement in follow up.

The issue where users can type a long heading but it gets clipped on blur is occurring because the latest changes to the TitleCellUpdateAttrsSchema weren’t reflected due to the shared package not being built. In our monorepo setup, the changes only take effect after the package has been properly built.

But I saw you did a great job on ux wise so this doesn't matter

benjreinhart:
> In our monorepo setup, the changes only take effect after the package has been properly built.

Yes I noticed that as well. I thought we had fixed this in #201 but it looks like we still need to build. FWIW you can run `pnpm run build` in repo root when developing.

Aswanth-c:
> > In our monorepo setup, the changes only take effect after the package has been properly built.
> 
> Yes I noticed that as well. I thought we had fixed this in #201 but it looks like we still need to build. FWIW you can run `pnpm run build` in repo root when developing.

This branch was behind the turbo branch. I think  running `turbo dev` on a newer versions  fix this issue

PR #222, opened 2024-08-27
Warnings, Underlines, & Hover Information for Issues

versecafe:
Goal to provide a smoother and more cohesive experience similar to working in IDE with proper hover tips, warnings, errors, and integrating these into srcbook in a clean style
 
- [X] Support TS warnings to frontend
- [X] AI Fix on TS warnings
- [x] Underlines in editor cell for warnings and errors

Splitting this off into a second PR
- Documentation on hover

### DIFF
--- a/packages/shared/src/types/tsserver.ts
+++ b/packages/shared/src/types/tsserver.ts
@@ -4,3 +4,4 @@ import { TsServerLocationSchema, TsServerDiagnosticSchema } from '../schemas/tss
 
 export type TsServerLocationType = z.infer<typeof TsServerLocationSchema>;
 export type TsServerDiagnosticType = z.infer<typeof TsServerDiagnosticSchema>;
+export type TsServerSuggestionType = TsServerDiagnosticType;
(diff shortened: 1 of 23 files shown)


versecafe:
image

versecafe:
@benjreinhart PR 1 of 2 covers the inline errors and hover, I'll make a second one for the on hover information from the LSP

versecafe:
image

versecafe:
image

packages/shared/src/types/tsserver.ts
```diff
@@ -4,3 +4,4 @@ import { TsServerLocationSchema, TsServerDiagnosticSchema } from '../schemas/tss
 
 export type TsServerLocationType = z.infer<typeof TsServerLocationSchema>;
 export type TsServerDiagnosticType = z.infer<typeof TsServerDiagnosticSchema>;
+export type TsServerSuggestionType = TsServerDiagnosticType;
```
benjreinhart, line 7: Even though types are the same today, perhaps it would be best to infer this from the corresponding schema rather than the other one?

benjreinhart (approved):
This is great, thank you!

Only noticed one minor issue, which is that it looks like the squigglies are sometimes off by one character

Screenshot 2024-08-27 at 4 19 12 PM

packages/shared/src/types/tsserver.ts
versecafe: Changed to independant schema and infer in case they every divulge

PR #226, opened 2024-08-28
[SPIKE] ui component-like interface

benjreinhart:
Spiking with some UI ideas. Eventually we will want to support JSX and react components inline. But we're also curious to explore something that looks like [Kino](https://github.com/livebook-dev/kino). This branch is to play with ideas in that space.

To try out this branch, you can download the branch and use [this Srcbook](https://github.com/user-attachments/files/16789987/ui-spike.src.md).

## Simple input example

Writing a code cell that looks like the following:

```typescript
import {input} from '@srcbook/ui';

const name = await input({label: 'Your name'});

console.log(`Hello, ${name}!`);
```

will dynamically render an `<input />` labeled `Your name` and wait for the user to enter input before proceeding.

### Waiting on user to enter value

Screenshot 2024-08-28 at 2 29 44 PM

### After entering a value

Screenshot 2024-08-28 at 2 30 59 PM

## Notes

There are all sorts of issues with this approach and how extensible / robust it is. For one, the example DSL is elegant using `await`, but in reality the input is likely part of a bigger form. How do we construct composite components? We can't always assume an input is used one time, so the `await` likely won't work (at least not in all cases). Do we then use some sort of callback that fires each the input changes? Do we fire on each change, or only on a submit? How do we update or remove elements from the server (e.g., remove form on submit and replace with some other component)? How do multi-step workflows work?


benjreinhart:
Closing because we're looking into supporting react components in Srcbook

PR #227, opened 2024-08-29
Hover tooltips for non errors/warnings

versecafe:
- [x] Basic hover UI
- [X] Cell based identification
- [x] TS Server description/types
- [x] TS Server docs
- [x] Beautify the tooltips!


versecafe:
image

Doesn't provide styling yet but it has usable hover docs!

benjreinhart (commented):
We're going to want to touch up the UI before merging, but functionality wise, this is looking great!

Basic UI touchups before merge:

1. We'll want some padding and probably a small border radius (3px is what we use in many cases)
2. Spacing between the sections (documentation / definition)
3. Cap the view to a certain height and width and having scrolling overflow
4. Probably a border color and maybe a shadow.

For these styles, if possible it would be great to use our tailwind classes. Given the imperative programming, this is annoying but worth it so we can maintain consistency in our styles.

packages/api/server/ws.mts
```diff
@@ -682,6 +686,47 @@ async function tsconfigUpdate(payload: TsConfigUpdatePayloadType) {
...
+
+  if (!session) {
+    throw new Error(`No session exists for session '${payload.sessionId}'`);
+  }
+
+  if (session.language !== 'typescript') {
+    throw new Error(`tsserver can only be used with TypeScript Srcbooks.`);
+  }
+
+  const tsserver = tsservers.has(session.id) ? tsservers.get(session.id) : createTsServer(session);
```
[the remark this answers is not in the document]
benjreinhart: Scratch that, I think I'm wrong here. Nothing would prevent users from hovering over code and sending this to the backend when a new srcbook is either created or imported before deps are installed (because tsserver wouldn't be available yet).

If deps are not yet installed then we don't want this to fail / break. We probably need to ignore failures here?

packages/api/server/ws.mts
```diff
@@ -682,6 +686,47 @@ async function tsconfigUpdate(payload: TsConfigUpdatePayloadType) {
...
+    line: payload.request.location.line,
+    offset: payload.request.location.offset,
+  });
+
+  const body = tsserverResponse.body;
+  if (!body) {
+    return null;
+  }
+
+  const response: TsServerQuickInfoResponseType = {
```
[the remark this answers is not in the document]
versecafe: yep, just a hold over from when the data was shaped differently, will fix

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+  }
+  // Add other quick info fields
+  if (info.quickInfo.displayString) {
+    const displayDiv = dom.appendChild(document.createElement('div'));
+    displayDiv.innerText = info.quickInfo.displayString;
+  }
+
+  if (info.quickInfo.kind) {
```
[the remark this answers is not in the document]
versecafe: Not sure about vscode but a subset of them is used often in Zed for TS, it usually gets reserved for methods on classes or aliases to make things clear when they're not obvious

packages/shared/src/schemas/tsserver.ts
```diff
@@ -20,3 +20,40 @@ export const TsServerSuggestionSchema = z.object({
...
+  displayString: z.string(),
+  documentation: z.union([
+    z.string(),
+    z.array(
+      z.object({
+        text: z.string(),
+        kind: z.string(),
+      }),
+    ),
+  ]),
```
versecafe, line 42: Move to it's own schema as it's used multiple times

versecafe:
continued UI progress
image

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,116 @@
...
+}
+
+import { EditorView, TooltipView } from '@uiw/react-codemirror';
+
+export type TooltipRenderer = (arg0: HoverInfo, editorView: EditorView) => TooltipView;
+
+export const hoverRenderer: TooltipRenderer = (info: HoverInfo) => {
+  console.log(info);
+  const dom = document.createElement('div');
+  dom.className = 'p-3 bg-background border border-border max-w-96 max-h-64 overflow-y-auto';
```
benjreinhart, line 77: I think we should make a couple changes here.

```
p-2 space-y-3 bg-background border border-border max-w-lg max-h-64 text-xs overflow-auto
```

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  if (info.quickInfo.displayString) {
+    const displayDiv = dom.appendChild(document.createElement('div'));
+    displayDiv.innerText = info.quickInfo.displayString;
+  }
+
+  if (info.quickInfo.documentation) {
+    for (const part of info.quickInfo.documentation) {
```
benjreinhart, line 85: I see `documentation` typed as:

```typescript
string | {
  text: string;
  kind: string;
}[] | undefined
```


In the examples I was using locally, documentation was a string, meaning this code was wrapping each character in the string with a `span`. I'm guessing that was not intended?

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,116 @@
...
+      const span = dom.appendChild(document.createElement('span'));
+      span.className = 'text-sb-core-50 pt-2';
```
benjreinhart, line 87: We have a class `text-tertiary-foreground` that is used for muting text a bit. This has variants for light and dark mode (which happen to be the same right now, but that could change).

packages/web/src/components/cells/hover.tsx
benjreinhart: Also I see that there are sometimes newlines added to the documentation string. It looks like vscode strips single newlines to flatten text for their display since the hover info otherwise can look funky with random newlines. See below, there's a newline after "result".

I wonder if we can find their logic for this or simply just strip single newlines?

Screenshot 2024-08-31 at 2 53 18 PM

benjreinhart:
I noticed view def and errors / suggestions on the same term also could use some consistency if possible

Screenshot 2024-08-31 at 2 56 56 PM

packages/web/src/components/cells/hover.tsx
benjreinhart: Hmm, maybe these actually:

```
p-2 space-y-3 bg-background dark:border dark:border-border shadow-lg max-w-lg max-h-64 text-xs overflow-auto
```

Adds a shadow, and applies border only in dark mode because for some reason in light mode the border looks prominent because I think codemirror is also adding one but you can't see it in dark mode? Not sure exactly. But I'd take a look in both light and dark mode.

versecafe:
image

versecafe:
@benjreinhart ready to go, could use some nice to haves for highlighting but thats going to take a lot of extra work this currently covers nice highlights, better UI on errors/warnings, docs, type, etc

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,143 @@
...
+  return hoverTooltip(async (view, pos) => {
+    if (cell.language !== 'typescript') {
+      return null; // bail early if not typescript
+    }
+
+    const { from, to, text } = view.state.doc.lineAt(pos);
+    let start = pos;
+    let end = pos;
+
+    while (start > from && /\w/.test(text[start - from - 1])) start--;
```
nichochar, line 24: Is this regex sufficient for all symbols that we might hover over?
(I think so... just checking)

packages/web/src/index.css
```diff
@@ -4,6 +4,23 @@
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
```
nichochar, line 7: I think there is a section lower in the file with other `.cm-X` stuff, can we move that there?

packages/shared/src/schemas/tsserver.ts
```diff
@@ -20,3 +20,34 @@ export const TsServerSuggestionSchema = z.object({
...
+    z.array(
+      z.object({
+        text: z.string(),
+        kind: z.string(),
+      }),
+    ),
+  ])
+  .optional();
+
+export const TsServerQuickInfoResponseSchema = z.object({
```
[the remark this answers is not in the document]
versecafe: I dredged up those fields from manual testing, finding actual documentation for the internal fields it has, I can't find any quickInfo specific docs to link to unfortunately

packages/web/src/components/cells/hover.tsx
versecafe: yep it's just for calculating where to snag the quickInfo and where to render the tooltip from, nothing else

packages/web/src/index.css
versecafe: @nichochar moved down with others

packages/web/src/index.css
```diff
@@ -392,3 +392,18 @@
     'Liberation Mono', 'Courier New', 'monospace';
   font-size: 14px;
 }
+
+.cm-tooltip {
```
benjreinhart, line 396: I think adding a box shadow here would be nice, particularly in light mode to give some depth.

Maybe `shadow-md` from tailwind?

```css
  box-shadow:
    0 4px 6px -1px rgb(0 0 0 / 0.1),
    0 2px 4px -2px rgb(0 0 0 / 0.1);
```

benjreinhart (commented):
Looking really good, there are a couple small things I'd still love to squeeze in here.

Did you consider trying to strip single newlines from doc strings? E.g.:

Screenshot 2024-09-01 at 2 06 25 PM

packages/web/src/components/cells/hover.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+  if (!hoverInfo) return null;
+
+  return (
+    <div className="p-2 space-y-3 max-w-lg max-h-64 text-xs overflow-auto relative">
+      {hoverInfo.displayString && <span>{hoverInfo.displayString}</span>}
+      {hoverInfo.documentation && (
+        <>
+          <br />
+          {typeof hoverInfo.documentation === 'string' ? (
```
benjreinhart, line 89: Do we want some padding or extra line between displayString and docs?

Screenshot 2024-09-01 at 2 18 35 PM

packages/web/src/index.css
```diff
@@ -392,3 +392,18 @@
...
+  background-color: hsl(var(--background)) !important;
+  border-color: hsl(var(--border)) !important;
+  border-width: 1px !important;
+  border-radius: 0.375rem !important;
```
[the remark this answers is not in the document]
versecafe: 0.375 if the md rounded size should match the rounding of the cells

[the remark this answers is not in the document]
benjreinhart: It should not match the cells, it should be less.

benjreinhart (approved):

benjreinhart:
woot this looks awesome, so excited for these! :shipit:

Issue #218, opened 2024-08-26
Add AI generation to markdown cells


nichochar:
Currently, the AI functionality allows users to:
- generate a new srcbook from a prompt
- insert new cells from a prompt
- iterate on a code cell with a prompt

One missing piece is the ability to edit a markdown cell with a prompt. The UX and mechanism should be quite similar to how we're doing it for code cells.

Aswanth-c:
Hey, I have started working on this, feel free to assign the task

Sanjoy-droid:
@nichochar is this issue still open or it is solved by now by @Aswanth-c

PR #228, opened 2024-08-29
Feat/#218 add ai generation to markdown cells

Aswanth-c:
- Moved Prompt section ui  to separate file `src/components/ai-prompt-input.tsx` for reuseablity 
- Added ai generation functionality to markdown 
- mentioned issue https://github.com/srcbookdev/srcbook/issues/218


benjreinhart (commented):
Thank you for this @Aswanth-c 

I pulled down the branch and am trying it out locally. Extracting the AI prompt looks like a good change, but the functionality does not seem to work as I would expect. The goal here is for the AI to generate markdown but the prompts submitted to the models are prompts that instruct the AI to write code. The result is that it seems to mostly generate code in the markdown cell which is not intended -- we want it to write markdown.

I believe we will need a new prompt for this PR.

Aswanth-c:
> Thank you for this @Aswanth-c
> 
> I pulled down the branch and am trying it out locally. Extracting the AI prompt looks like a good change, but the functionality does not seem to work as I would expect. The goal here is for the AI to generate markdown but the prompts submitted to the models are prompts that instruct the AI to write code. The result is that it seems to mostly generate code in the markdown cell which is not intended -- we want it to write markdown.
> 
> I believe we will need a new prompt for this PR.



> Thank you for this @Aswanth-c
> 
> I pulled down the branch and am trying it out locally. Extracting the AI prompt looks like a good change, but the functionality does not seem to work as I would expect. The goal here is for the AI to generate markdown but the prompts submitted to the models are prompts that instruct the AI to write code. The result is that it seems to mostly generate code in the markdown cell which is not intended -- we want it to write markdown.
> 
> I believe we will need a new prompt for this PR.

Gotcha, I will update the pr with new prompt

Aswanth-c:
> > Thank you for this @Aswanth-c
> 
> > 
> 
> > I pulled down the branch and am trying it out locally. Extracting the AI prompt looks like a good change, but the functionality does not seem to work as I would expect. The goal here is for the AI to generate markdown but the prompts submitted to the models are prompts that instruct the AI to write code. The result is that it seems to mostly generate code in the markdown cell which is not intended -- we want it to write markdown.
> 
> > 
> 
> > I believe we will need a new prompt for this PR.
> 
> 
> 
> 
> 
> 
> 
> > Thank you for this @Aswanth-c
> 
> > 
> 
> > I pulled down the branch and am trying it out locally. Extracting the AI prompt looks like a good change, but the functionality does not seem to work as I would expect. The goal here is for the AI to generate markdown but the prompts submitted to the models are prompts that instruct the AI to write code. The result is that it seems to mostly generate code in the markdown cell which is not intended -- we want it to write markdown.
> 
> > 
> 
> > I believe we will need a new prompt for this PR.
> 
> 
> 
> Gotcha, I will update the pr with new prompt 

Update: i have added prompt,
will do through testing 
Expect the pull request to be ready for review  by tommorow

versecafe (approved):
@benjreinhart Works for me and looks good but could you take a look at it, and make sure the prompt is good

packages/api/prompts/code-updater-markdown.txt
```diff
@@ -0,0 +1,85 @@
...
+
+## Final Instructions
+
+The user's Srcbook will be passed to you, surrounded with `==== BEGIN SRCBOOK ====` and `==== END SRCBOOK ====`.  
+The specific **Markdown cell** they want updated will also be passed to you, surrounded with `==== BEGIN MARKDOWN CELL ====` and `==== END MARKDOWN CELL ====`.  
+The user's intent will be passed to you between `==== BEGIN USER REQUEST ====` and `==== END USER REQUEST ====`.
+
+Your job is to edit the cell based on the contents of the Srcbook and the user's intent.  
+Act as a **Markdown expert**, writing the best possible content you can. Focus on being **elegant, concise, and clear**.  
+**ONLY RETURN THE MARKDOWN TEXT , NO PREAMBULE, NO SUFFIX, NO CODE FENCES (LIKE TRIPLE BACKTICKS) ONLY THE MARKDOWN**.
```
whatrocks, line 85: `PREAMBULE` should be `PREAMBLE`

packages/api/prompts/code-updater-markdown.txt
```diff
@@ -0,0 +1,85 @@
...
+A Srcbook is a **Markdown-compatible notebook**, used for documentation or text-based content.
+
+### Srcbook Spec
+
+The structure of a Srcbook:
+0. The language comment: `<!-- srcbook:{"language":"markdown"} -->`
+1. Title cell (heading 1)
+2. N more cells, which are either:
+   - **Markdown cells** (GitHub flavored Markdown)
+   -  Markdown cells, which have a filename and source content.
```
nichochar, line 16: This is wrong, should be code cells

nichochar (commented):
There's a few issues with this and I think it's not worth fixing them as we're changing how the AI works (see app builder branch)

PR #230, opened 2024-08-30
[SPIKE] UI Rendering with external @srcbook/ui-sdk lib + websockets transport

nichochar:
This MVP attempts to create a UI app interactively within Srcbook, but does it by running a cell, which declaratively describes a UI app at a high level of functionality. The SDK is inspired by interval.com's one, designed to build very opinionated internal tooling, inter-twined with node backend code.

The `@srcbook/api` is now a "proxy" websocket server, which broadcasts events back to the `session:<id>` channel. This is because otherwise registration between the node process running `@srcbook/ui-sdk` and the frontend would rely on the server anyways. Down the line, we can likely create better channel hygiene and not re-broadcast all events like I do now.

Here is the TS cell creating the app:
```typescript
import { Action, io, UIApp } from '@srcbook/ui-sdk';

const greetingAction = new Action('greeting', async () => {
  const name = await io.input.text("What's your name?");
  const age = await io.input.text("How old are you?");
  const greeting = `### Hello, ${name}!\n\nYou are \`${age}\` years old.`;
  await io.display.markdown(greeting);
  return greeting;
});

// UIApp knows how to connect to the proxy WS server
// through an env var passed to it
const app = new UIApp(greetingAction);
app.run();
```

## Gif of experience (no design effort made)
CleanShot 2024-08-29 at 20 43 17

## Events

Currently, the node process running in the cell sends a serialized component spec with an `ui:io:await_response` event. The client renders this, and once the user submits the interaction, the web client sends a `ui:io:response` event which leads to a promise getting resolved, which leads to the code continuing it's execution.

## Ideas that I will explore further
- Need to try more complex DSL things. In particular grouping of fields
- Need to introduce "start" "stop" mechanics for the app
- Need to support cancellation flows
- Rethink the events, which are very simple right now.


nichochar:
We are closing this SPIKE for now. We like the direction but want to explore another alternative of rendering browser cells (HTML, .jsx, react apps) and see how that feels also.

The long term goal is to help developers build and deploy entire apps, and the UI kit might be limiting for more complex or unique apps. It might still have a place on the "fast and efficient" side of the tradeoff, but we'll revive it at that point if the time comes. Keeping the branch alive.

Issue #209, opened 2024-08-24
Support formatting (prettier?)


nichochar:
### Objective
Integrate Prettier (or alternative, maybe biome?) for code formatting in our TypeScript notebook app.

### Requirements
1. Implement Prettier formatting for TypeScript cells
2. Add auto-format option in app settings
3. Create a format button in the cell action bar, for doing it manually on a per-cell basis

### Implementation Steps
1. Install Prettier and necessary plugins to default package.json
2. Integrate Prettier with CodeMirror editor
3. Implement formatting function for individual cells
4. Add auto-format toggle to app settings
5. Design and implement format button in cell action bar

Issue #210, opened 2024-08-24
Auto-install deps when creating a new TS Srcbook


nichochar:
### Objective
Automatically install required dependencies (tsx, typescript, @types/node) when creating a new TypeScript Srcbook, eliminating the need for the "install dependencies" toast, which needs to be clicked 100% of the time.

### Current Behavior
- "Install dependencies" toast appears when creating a new TS Srcbook

### Proposed Change
- Automatically install tsx, typescript, and @types/node upon TS Srcbook creation

### Considerations
- Block running cells until completed?
- Handle scenarios where auto-installation fails
- Consider allowing users to opt-out of auto-installation

swk777:
I’m considering adding a flag in the router, such as isAutoInstallDeps, to enable automatic installation of the necessary dependencies when creating a new TypeScript Srcbook or opening other Srcbooks if needed.
Would this approach work for you?

destinedcodes:
can I work on this??

just found out srcbook today and it's just been amazing. I'd like to make some contributions

nichochar:
Hey @destinedcodes, sounds good, I've assigned the ticket to you. If you want to discuss the implementation or have any issues on the dev side of Srcbook, we have a [Discord](https://discord.gg/shDEGBSe2d)

BeRecursive22:
@nichochar @benjreinhart I was observing a weird thing with the prettier formatting. The cell is not executed and there's no output on stdout when you create a new code cell, write some typescript and then hit enter. But the same cell is executed without any troubles if we format it before running it. 

So I'm just pushing one more event (cell:format) right after cell:create and this takes care of the dependency thing. 

<img width="649" alt="image" src="https://github.com/user-attachments/assets/8b826fc6-eca6-47e2-9f26-4af810385b56">

[screen-capture (16).webm](https://github.com/user-attachments/assets/d98ec8f7-937f-4736-a7c4-402028fb4f88)


With above approach, the "Need to install deps" toast would not make much sense since the cell is ready to be executed even before having an actual code.

Aswanth-c:
Hey I'm back in action, started working on it, 
feel free to assign 

> Consider allowing users to opt out of auto-installation
for this, I'm considering a toggle option in the settings. Let me what you think about this approach.

Aryann15:
Hey @nichochar, Can I work on this issue?

nichochar:
>for this, I'm considering a toggle option in the settings. Let me what you think about this approach.

[...]

PR #231, opened 2024-08-30
Feat: Add Prettier Support to TypeScript Notebook

swk777:
implement  #209 

- [x] Implement server-side code formatting using Prettier
- [x] Add a button and hotkey for triggering code formatting
- [x] Support customizable Prettier configurations

**Minor Details**
      • Upgraded Radix UI to include the brush icon
      •  Moved Prettier configuration to package.json since having a standalone configuration doesn’t seem crucial at this point.
      • Set Command+Shift+F as the shortcut for formatting.


versecafe:
Should this be debounced auto or a manual format button & key bind to avoid changing the lines around as someone works if they pause between typing?

versecafe:
Could this be done on the server to avoid making the client bundle even heavier and avoid the need to send the entire formatted file down to the server, change the fs, and then sync back with the client

swk777:
> Could this be done on the server to avoid making the client bundle even heavier and avoid the need to send the entire formatted file down to the server, change the fs, and then sync back with the client

It’s true, but the lengthy process isn’t ideal for auto-formatting, and the sync model is currently not supported.

packages/web/src/components/cells/code.tsx
```diff
@@ -434,6 +453,22 @@ function Header(props: {
...
+              <Tooltip>
+                <TooltipTrigger asChild>
+                  <Button
+                    variant="icon"
+                    size="icon"
+                    disabled={cellMode !== 'off'}
+                    onClick={onFormatCode}
+                    tabIndex={1}
+                  >
+                    <CodeXml size={16} />
```
benjreinhart, line 466: Couple notes on this:

* I wonder if there's a common icon for code formatting? This icon is ambiguous and could be used to indicate many different things in a product like ours. Maybe [braces](https://lucide.dev/icons/braces), but that doesn't feel much different?
* This icon feels a bit odd with the current debounce UX. You have < 2 seconds after typing anything before this pressing this icon will have no affect. That said, it is useful if you import a srcbook and want to format the cell without typing in it first.

packages/web/src/components/cells/code.tsx
```diff
@@ -643,6 +680,27 @@ function CodeEditor({
 
   const updateCellOnServerDebounced = useDebouncedCallback(updateCellOnServer, DEBOUNCE_DELAY);
 
+  const formatCellCode = async (source: string, viewUpdate: ViewUpdate) => {
```
benjreinhart, line 683: Nit: There's no `await` inside this function so we don't need the `async` keyword in this case. Alternatively can change to using await and ditch the promise callbacks.

benjreinhart (commented):
Thank you for this!

TLDR: implementation looks great but I'm not sure this is the UX we want.

* I think having code shift out from under me after a slight pause in typing is not ideal. In my editor it formats only on save which is an explicit action I take. In Srcbook, we don't explicitly save, so that won't work here. However, I think we can apply the same principle and use a keyboard shortcut and/or icon as the explicit action for formatting. This way I opt-in to each format which I feel pretty strongly about.
* If we use a keyboard shortcut and/or icon, then we don't need the database entry for auto format.
* I'm torn on the client-side vs server-side formatting functionality. I would prefer it to take place inside the folder we create for the Srcbook (similar to how we handle tsconfig and type checking). That is, I would expect that repository to have prettier installed in its deps and the prettier config to be located in that folder as well. In theory, this makes ejecting a srcbook folder a breeze, which is something I think we'll eventually want to do. BUT, if the user presses the format icon and there's a delay I agree that's a poor experience. I wonder if it's fast enough there's no real delay?
* It's ok if we do it in a follow up PR, but the prettier settings should be editable by the user, the same as package.json or tsconfig.json.

Excellent work on implementation here but I think we should change the UX

versecafe:
@swk777 @benjreinhart On the server driven delay the back and forth with the WS in local is ~50ms (hardware dependent) under the current system including validation with zod, so it should not create any sort of noticeable delay to a user, this may be different for a hosted instance but even a 500ms delay if you play a little animation isn't an issue since the format is something you do maybe once or twice when working on a cell.

packages/web/src/components/cells/code.tsx
nichochar: I like, and I think I've seen this before, using a brush icon:
CleanShot 2024-09-01 at 13 33 43

`PaintbrushVertical` in lucide-react.

nichochar (commented):
Thanks for the contrib @Aswanth-c 

I apologize, I think the ticket should have been more specific on what we wanted, but I agree with Ben's review. I think a better behavior is to implement the formatting on the backend side, with the ability to trigger it from the frontend. So:
- on each cell, you can click a button in the action bar (`PaintbrushVertical`) which formats it by calling a websocket event `CellUpdateFormatCell` and updating the cell.source.
- [Stretch 1] Auto-format on export: if auto-format is enabled, we format all saves _on export only_. Otherwise as @benjreinhart says, things change while you're typing and that's a very jarring experience.
- [Stretch 2]: add a global keyboard shortcut to format all cells (and/or format current active cell).

swk777:
Thank you for the detailed feedback

1.	I’ll dig into implementing the formatting on the server side as suggested.
2.	To keep the scope of this PR manageable, I’d like to start by implementing formatting for individual cells, triggered by a button or keyboard shortcut. We can then use this for a while to see if there’s a real need for global formatting. If it turns out to be necessary, we can easily add it later, as the implementation should be straightforward.

Does this approach sound good to you? @benjreinhart @nichochar

nichochar:
Approach sounds good to me. I would add a button with the paintbrush right here to start:
CleanShot 2024-09-01 at 20 14 01

versecafe:
@swk777 When you're finishing this up can you merge main back in and run `pnpm changeset` to create the changelog

versecafe:
Should format errors be piped up out to be more visible as the current toast doesn't explain the error but it is nicely captured in the console as 

```
   21 |  */
   22 | export class TicTacToe {
 > 23 |   private board: string[[];
      |                           ^
   24 |   private currentPlayer: string;
   25 |
   26 |   constructor() {
```
image

versecafe (approved):
Code side all looks good just the note on making errors on format runs include more information on the front end.

swk777:
> Code side all looks good just the note on making errors on format runs include more information on the front end.

It’s true, I’ve now routed the output to stderr. The error does overlap somewhat with the ones in the problems panel, but it currently depends on the TypeScript package installation. Once #210  is completed, I can work on further optimizations.

packages/api/session.mts
```diff
@@ -296,7 +298,69 @@ export function updateCell(session: SessionType, cell: CellType, updates: CellUp
       return updateCodeCell(session, cell, updates);
   }
 }
-
+async function ensurePrettierInstalled(dir: string): Promise<boolean> {
```
benjreinhart, line 301: Some thoughts here:

1. If formatting is going to be a first class citizen in all Srcbooks, then we should create every new srcbook with it already present in the package.json's `devDependencies`
2. If we do check for its presence here, are we able to use [`depcheck`](https://github.com/depcheck/depcheck)? We already use this in `deps.ts` and it looks like it has "special" support for prettier.
3. If the dep is missing, I'm inclined to prompt the user for installing it. While this does introduce an extra manual step, it keeps the user in the loop and gives faster feedback than if they waited for install to happen behind the scenes. Most importantly, this should rarely happen if we include prettier in the dependencies when creating new srcbooks.

packages/api/session.mts
```diff
@@ -296,7 +298,69 @@ export function updateCell(session: SessionType, cell: CellType, updates: CellUp
...
+  try {
+    // check if prettier is installed
+    await fs.access(prettierPath);
+    return true;
+  } catch (error) {
+    return new Promise<boolean>((resolve) => {
+      try {
+        npmInstall({
+          cwd: dir,
+          packages: ['prettier'],
```
benjreinhart, line 312: We should install this in the devDependencies

benjreinhart (commented):
This is looking really good, excited for this. Couple last minute items but it's almost good to go. Thank you!

packages/api/session.mts
benjreinhart: I stand behind number 3 above but in the interest of getting this out sooner than later, we can ignore that piece for now and leave this code for now (though would be good to use depcheck).

benjreinhart: Ohh I just noticed we are doing #1 here already, nice

packages/web/src/components/cells/code.tsx
```diff
@@ -660,7 +715,18 @@ function CodeEditor({
     javascript({ typescript: true }),
     tsHover(session.id, cell, channel, theme),
     tsLinter(cell, getTsServerDiagnostics, getTsServerSuggestions),
-    Prec.highest(keymap.of([{ key: 'Mod-Enter', run: evaluateModEnter }])),
+    Prec.highest(
+      keymap.of([
+        { key: 'Mod-Enter', run: evaluateModEnter },
+        {
+          key: 'Mod-Shift-f',
```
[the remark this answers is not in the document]
versecafe: This just got me thinking should we add srcbook keymap support as a JSON and with the default, vscode, jetbrains, vim keymaps vim may be hard for nav but at least for basic actions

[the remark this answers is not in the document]
benjreinhart: That would be cool but I would say that's out of scope for this PR (not that you're proposing doing it here)

benjreinhart:
You're going to want to rebase main as well given the navigation updates (#267).

swk777:
all done, thx

.changeset/thin-dots-rest.md
```diff
@@ -0,0 +1,7 @@
+---
+'@srcbook/shared': minor
```
benjreinhart, line 2: I also just noticed this is `minor`, but for now I would like to keep all the versions on `patch`. Do you mind regenerating this changeset such that these are all `patch`?

versecafe: you don't need to regen just change the text to patch and it'll work

benjreinhart:
This is awesome, appreciate all your work here! 🙏🏻

nichochar:
Amazing, thanks for the great contribution, really like having it in there, it feels great. After using it a couple of times, I think we'll want to add some indication that it's running (like AI has a purple ring, and running a process has a yellow ring), but we'll ask design for this.

Thanks @swk777 🚀 👊

PR #238, opened 2024-08-31
Adds eslint to CI and supports _ignore

### LINKED ISSUE #236: CI doesn't seem to run FE linter
CI used to run eslint and catch things like missing dependencies in useEffect. I think it may have gotten dropped in #201 

cc @versecafe

versecafe:
Fixes #236

### DIFF
--- a/packages/web/.eslintrc.cjs
+++ b/packages/web/.eslintrc.cjs
@@ -12,5 +12,13 @@ module.exports = {
   plugins: ['react-refresh', 'prettier', 'jest'],
   rules: {
     'prettier/prettier': 'error',
+    '@typescript-eslint/no-unused-vars': [
+      'warn', // or "error"
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+        caughtErrorsIgnorePattern: '^_',
+      },
+    ],
   },
 };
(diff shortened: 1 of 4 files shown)


packages/web/.eslintrc.cjs
```diff
@@ -12,5 +12,13 @@ module.exports = {
   plugins: ['react-refresh', 'prettier', 'jest'],
   rules: {
     'prettier/prettier': 'error',
+    '@typescript-eslint/no-unused-vars': [
+      'warn', // or "error"
```
benjreinhart, line 16: This still means that we will fail on unused variables, just not when they start with the below patterns, correct?

benjreinhart (approved):

packages/web/.eslintrc.cjs
versecafe: Correct, because max-warnings is set to zero it will still fail a build but will display as a warning instead of an error, but if you want it to also display as an error I can flip it over

benjreinhart: As long as it fails the build (and doesn't start with a `_`), I don't care if it's warning or error.

benjreinhart:
🙏🏻

packages/web/.eslintrc.cjs
versecafe: yep build always fails if an unused var isn't prefixed with `_name` or just named `_`

PR #240, opened 2024-09-01
removes between 1-6 layers of uneeded div nesting

versecafe:
Just a cleanup PR pulling out several dozen instances of uneeded excessive div nesting for styling by directly applying styles to components, makes dom faster to parse, bit easier for screen readers, and just easier to navigate through while debugging.

### DIFF
--- a/packages/web/index.html
+++ b/packages/web/index.html
@@ -29,7 +29,7 @@
     </script>
   </head>
   <body>
-    <div id="root"></div>
+    <div id="root" class="flex flex-col"></div>
     <script type="module" src="/src/main.tsx"></script>
   </body>
 </html>
--- a/packages/web/src/components/cells/code.tsx
+++ b/packages/web/src/components/cells/code.tsx
@@ -187,7 +187,7 @@ export default function CodeCell(props: {
   }
 
   return (
-    <div id={`cell-${props.cell.id}`}>
+    <>
       <Dialog open={fullscreen} onOpenChange={setFullscreen}>
         <DialogContent
           className={cn(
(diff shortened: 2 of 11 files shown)


packages/web/src/components/cells/title.tsx
```diff
@@ -15,8 +15,11 @@ export default function TitleCell(props: {
   }
 
   return (
-    <div id={`cell-${cell.id}`} className="mb-4">
-      <EditableH1 text={cell.text} className="title" onUpdated={updateCell} />
-    </div>
+    <EditableH1
+      text={cell.text}
+      id={`cell-title-${cell.id}`}
```
benjreinhart, line 20: `cell-${cell.id}` is used to target this element for scrollTo purposes, so I believe this change will break that behavior?

versecafe: prior to this change there was multiple cell-${cell.id} instances which was doing weird things

packages/web/index.html
```diff
@@ -29,7 +29,7 @@
     </script>
   </head>
   <body>
-    <div id="root"></div>
+    <div id="root" class="flex flex-col"></div>
```
nichochar, line 32: Agree with most of these changes, but I prefer we have 0 styling in this file. It makes for a cleaner root and need to open less files when "looking for the rule".

I would add this back to the Layout

packages/web/src/components/cells/code.tsx
```diff
@@ -253,66 +253,64 @@ export default function CodeCell(props: {
...
-                fixDiagnostics={aiFixDiagnostics}
-                cellMode={cellMode}
-                fullscreen={fullscreen}
-                setFullscreen={setFullscreen}
-              />
-            </>
-          )}
-        </div>
+      <div
+        className={cn(
```
nichochar, line 257: I haven't tried to refactor this yet but I suspect we can improve states here as well.

packages/web/src/components/settings-sheet.tsx
```diff
@@ -113,18 +113,17 @@ function PackageJson({ openDepsInstallModal }: { openDepsInstallModal: () => voi
...
-            theme={codeTheme}
-            extensions={[
-              json(),
-              Prec.highest(keymap.of([{ key: 'Mod-Enter', run: evaluateModEnter }])),
-            ]}
-            onChange={onChangeSource}
-            basicSetup={{ lineNumbers: false, foldGutter: false }}
-          />
-        </div>
+        <CodeMirror
```
nichochar, line 116: I'm 50/50 on the readability improvement on this one. A surrounding div with only styles is easier to see than one of many props. Don't feel super strongly.

packages/web/src/components/settings-sheet.tsx
```diff
@@ -240,7 +238,7 @@ function Error(props: { error: string | null }) {
     <div className="px-1.5 pb-1.5">
       <div className="bg-error text-error-foreground text-sm font-medium py-2 pl-[10px] pr-3 flex items-start gap-1.5 rounded-sm">
         <div className="shrink-0 mt-0.5">
-          <Info size={16} />
+          <Info size={16} className="shrink-0 mt-0.5" />
```
nichochar, line 241: Did we improve anything here? We should either remove the div around it or not do this? 

Make sure to test this one, I'm not sure it's equivalent behavior (all for it if it is)

packages/web/src/components/cells/title.tsx
benjreinhart: the ids are unique so this shouldn't be problematic. What were the weird things that were happening?

benjreinhart:
@versecafe should we close this for now given the large set of view changes in #267 ?

PR #244, opened 2024-09-03
Syntax highlighting in hover tooltips

benjreinhart:
This adds syntax highlighting and formatting for the type signature + JS docs when hovering over a term in our code editor.

### Dark mode

Screenshot 2024-09-04 at 4 46 21 PM

### Light mode

Screenshot 2024-09-04 at 4 46 10 PM

### Notes

* This is far from perfect but it's an improvement on what's already there. The type signature is in a good place but the JS docs are styled using our existing markdown styles (`sb-prose` class) which is intended for blog-post like prose, not condensed tooltips. We will have to address this later.
* I removed react for now but only because this felt simple enough. That may be a bad decision though and may want to reintroduce, we'll see.

### DIFF
--- a/packages/web/src/components/cells/hover.tsx
+++ b/packages/web/src/components/cells/hover.tsx
@@ -1,24 +1,22 @@
-import {
+import type {
   CodeCellType,
-  TsServerQuickInfoRequestPayloadType,
+  TsServerJsDocTagsType,
+  TsServerJSDocType,
   TsServerQuickInfoResponsePayloadType,
-  TsServerQuickInfoResponseType,
 } from '@srcbook/shared';
-import { Extension, hoverTooltip, EditorView, EditorState } from '@uiw/react-codemirror';
+import { Extension, hoverTooltip } from '@uiw/react-codemirror';
 import { mapCMLocationToTsServer } from './util';
 import { SessionChannel } from '@/clients/websocket';
-import { createRoot } from 'react-dom/client';
-import { useEffect, useState } from 'react';
-import { javascript } from '@codemirror/lang-javascript';
 import { parse } from 'marked';
-import { type ThemeExtensionType } from '@/components/use-theme';
+import { formatCode } from '@/lib/code-theme';
+import { type ThemeType } from '@/components/use-theme';
 
 /** Hover extension for TS server information */
 export function tsHover(
   sessionId: string,
   cell: CodeCellType,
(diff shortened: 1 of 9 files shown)


packages/web/src/components/cells/hover.tsx
```diff
@@ -27,113 +33,102 @@ export function tsHover(sessionId: string, cell: CodeCellType, channel: SessionC
...
-          cell,
-          channel,
-          pos,
-        }),
+      create: () => {
+        const tooltipContainer = document.createElement('div');
+        tooltipContainer.className = 'p-2 space-y-2 max-w-3xl max-h-96 overflow-scroll';
+
+        function callback({ response }: TsServerQuickInfoResponsePayloadType) {
+          console.log(response);
```
versecafe, line 41: Probably shouldn't be in here

packages/web/src/index.css
```diff
@@ -394,19 +394,19 @@
...
-  border-radius: 0.1875rem;
-  overflow: hidden;
-  box-shadow:
-    0 4px 6px -1px rgb(0 0 0 / 0.1),
-    0 2px 4px -2px rgb(0 0 0 / 0.1);
+  @apply overflow-hidden;
+  @apply !bg-background;
+  @apply border;
+  @apply rounded-sm;
+  @apply shadow-md;
```
versecafe, line 401: Just a warning this will break if there are no other shadow-md uses in the code at some point since tailwind won't make the class

versecafe (commented):
Not an approval or change request up to Nick just things that could be improved or cause issue but don't need to be fixed

packages/web/src/index.css
benjreinhart: I think it should be scanning our css files but admittedly haven't verified. `.css` is included in [our tailwind config](https://github.com/srcbookdev/srcbook/blob/main/packages/web/tailwind.config.js#L6) though.

packages/web/src/components/cells/hover.tsx
benjreinhart: yes, you are right!

packages/web/src/index.css
versecafe: good question, I've never seen it used inside of the CSS file itself, worse case it breaks we search for the issue and the note will be here

versecafe (approved):

PR #247, opened 2024-09-03
Overhaul repo config handling

versecafe:


srcbook/package.json
```diff
@@ -5,17 +5,19 @@
   "type": "module",
   "bin": "./bin/cli.mjs",
```
benjreinhart, line 6: I think this is what is needed for the binary in production, not the `start` command below. This will need the proper path.

benjreinhart (approved):
sweet :shipit:

srcbook/package.json
versecafe: @benjreinhart yep totally missed that, fixed it and tested with `npm link` `srcbook start` works

PR #251, opened 2024-09-04
Update CONTRIBUTING.md to require pnpm version 9.5+

1egoman:
This was required when I set things up locally - I was running an older version of `9`, but that didn't support the `catalog:` stuff in some of the package.json's which led to install issues.


benjreinhart (approved):

1egoman:
Thanks!

PR #253, opened 2024-09-04
Rollout changesets

versecafe:
Just the changeset itself, release action + changeset action in separate PR to keep small and clean, the large line change is just the lock file


CONTRIBUTING.md
```diff
@@ -40,6 +40,26 @@ To run a script defined in one of the packages:
...
 
+## Making a changeset
+
+For creating release notes whenever a change is made
+
+```shell
+pnpm changeset
+```
+
+then follow through the CLI, provide a clean descriptionto your changes
```
nichochar, line 51: /s/descriptionto/description to/

srcbook/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "srcbook",
-  "version": "0.0.1-alpha.18",
```
nichochar, line 3: Can we keep the alpha here? Are we planning on fixing this as a follow up? I don't like the idea of releasing 0.0.1 by accident

benjreinhart: I'd prefer to keep alpha for a little longer but I do think we're close to 0.0.1 and then from there a long journey to 1.0.0.

benjreinhart (approved):

srcbook/package.json
versecafe: Ah yep let me put it back changeset doesn't really work with the alpha stuff since it's not semver compliant

PR #255, opened 2024-09-05
Socials in readme

nichochar:


README.md
```diff
@@ -7,6 +7,12 @@
 [![npm version](https://badge.fury.io/js/srcbook.svg)](https://badge.fury.io/js/srcbook)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
+<p align="center">
```
benjreinhart, line 10: If we're going to center these, can we also center the badges above? It looks off otherwise

PR #256, opened 2024-09-05
release CI

versecafe:
## Features
- Auto gens PR in sync with main to create change logs off changesets
- Creates github releases on PR merge
- Creates NPM releases on PR merge

## Prereq

- [x] NPM_TOKEN added to secrets
- [x] Allow GitHub Actions to create and approve pull requests set True

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,42 @@
+name: Release
+
+on:
+  push:
+    branches:
+      - main
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  release:
+    permissions:
+      contents: write
+      packages: write
+      pull-requests: write
+    name: Releasing
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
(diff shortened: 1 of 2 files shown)


benjreinhart:
How do we trigger a release, what is the process? Might be worth documenting this in the Contrib doc too?

versecafe:
image

benjreinhart:
When does the PR first open? What happens if something goes wrong in the process? Are there any docs I should at?

versecafe:
The PR opens and updates on all commits to main, just keeps one synced PR, I think it also has a prerequisite that a version actually changed, it's built on top of changesets@v1 https://github.com/changesets/action

benjreinhart (approved):
I think a small addition to the contributing.md doc that explains the flow of create changeset -> release works would be helpful (and also what to do if something fails perhaps)

versecafe:
> I think a small addition to the contributing.md doc that explains the flow of create changeset -> release works would be helpful (and also what to do if something fails perhaps)

Will add as long as everything goes smoothly with tomorrows test run of it

PR #259, opened 2024-09-06
patch empty tooltip bug

versecafe:


packages/web/src/components/cells/hover.tsx
```diff
@@ -35,9 +35,11 @@ export function tsHover(
       end: end,
       create: () => {
         const tooltipContainer = document.createElement('div');
-        tooltipContainer.className = 'p-2 space-y-2 max-w-3xl max-h-96 overflow-scroll';
+        tooltipContainer.style.display = 'none';
```
benjreinhart, line 38: Are we able to use tailwind for this property too?

benjreinhart (approved):

packages/web/src/components/cells/hover.tsx
versecafe: hidden or invisible might work similarly but they're not the same as none I can try a swap in if you want

benjreinhart: `hidden` translates to `display: none`.

Screenshot 2024-09-06 at 12 37 50 PM

versecafe: thought hidden had some extra on it since it breaks with flex, maybe none breaks with flex, I'll swap it

PR #264, opened 2024-09-07
Add a mailing list subscription upsell on front-page and settings

nichochar:
We want to collect users emails if they're interested so that we can send them updates, questionnaires, and tips. We want to talk to our users!

This adds a modal that pops up on the homepage. When the user either dismisses it or enters their email, it won't appear again.
The user can still go in the settings to add or modify their email subscription settings.

The feature is powered by sqlite.

The API supports unsubscribe, but I'd rather have that flow happen through the email newsletter than in the UI, so left it out.

Experience videos:

https://github.com/user-attachments/assets/f025f880-bd70-47ca-bddf-9f371c4ccc34

https://github.com/user-attachments/assets/f035f66a-8326-4c44-9c74-15b45b926c34

### DIFF
--- a/packages/web/src/routes/settings.tsx
+++ b/packages/web/src/routes/settings.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react';
 import { CircleCheck, Loader2, CircleX } from 'lucide-react';
-import { disk, updateConfig, aiHealthcheck } from '@/lib/server';
+import { disk, updateConfig, aiHealthcheck, subscribeToMailingList } from '@/lib/server';
 import { useSettings } from '@/components/use-settings';
 import { AiProviderType, getDefaultModel, type CodeLanguageType } from '@srcbook/shared';
 import type { SettingsType, FsObjectResultType } from '@/types';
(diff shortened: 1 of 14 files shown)


nichochar:
Added deletion API in the hub, can hook it up to add an unsubscribe: https://github.com/srcbookdev/hub/pull/4#issuecomment-2335027347

packages/web/src/routes/settings.tsx
```diff
@@ -211,6 +224,53 @@ function Settings() {
...
+              <Button
+                className="px-5"
+                onClick={async () => {
+                  try {
+                    const response = await subscribeToMailingList(email);
+                    if (response.success) {
+                      setSubscribedEmailStatus(email);
+                      setEmailStatus({ email, dismissed: false });
+                      toast.success('Subscribed successfully!');
+                    } else if (response.status === 409) {
```
benjreinhart, line 251: It'd be better to return success from the API when the user is already subscribed and then do away with this state which doesn't add anything IMO. This is especially true in this case because we hide the form afterwords and so they shouldn't typically have the ability to even do this.

benjreinhart (commented):
We should be using sqlite for the dismissed state not local storage.

nichochar:
> We should be using sqlite for the dismissed state not local storage

Can you share the reasoning?

packages/web/src/routes/settings.tsx
nichochar: It's slightly more info (you're already on there) but happy to simplify I like simple.

nichochar: Done

nichochar:
Updated to sqlite. Reasoning was "Really want to avoid spamming when users open in a private browser tab, and sqlite will handle those edge-cases"

benjreinhart (approved):

PR #267, opened 2024-09-09
Implement navigation changes

1egoman:
This change implements a new global navbar and sidebar, based on the designs in figma.

This is what the main session page (`/srcbooks/:id`) looks like now:
Screenshot 2024-09-09 at 4 14 01 PM

When a user clicks on any of the sidebar icons and the screen is thinner, a sheet opens:
Screenshot 2024-09-09 at 4 14 09 PM

When a user clicks on any of the sidebar icons and the screen is wider, a persistently open panel is shown:
Screenshot 2024-09-09 at 4 14 19 PM

Finally, the navbar carries over to the home page in addition to the session page:
Screenshot 2024-09-09 at 4 14 28 PM


packages/web/src/main.tsx
```diff
@@ -40,17 +43,28 @@ const router = createBrowserRouter([
...
+            path: '/settings',
+            element: <Settings />,
+            loader: Settings.loader,
+            action: Settings.action,
+            errorElement: <ErrorPage />,
+          },
+        ],
       },
     ],
   },
```
1egoman, line 70: I've had to make some relatively substantial changes to the root router at the top of the app - this is because I split the navbar up into two different underlying implementations, one that renders on the session page, and one that renders everywhere else, and the two different implementations require different data (which meant I had to use two different `loaders` depending on the navbar).

There may be a more elegant way to do this, but what I have here seems to work.

packages/web/src/routes/session.tsx
```diff
@@ -12,68 +12,105 @@ import {
...
-    return () => {
-      channel.unsubscribe();
-
-      if (session.language === 'typescript') {
-        channel.push('tsserver:stop', { sessionId: session.id });
-      }
-    };
-  });
+    forceComponentRerender();
+  }, [session.id, session.language, forceComponentRerender]);
```
1egoman, line 94: I had to make some signifigant changes to the way that the session page connected to the underlying websocket - the way it previously worked, if a user navigated internally within the app to a new srcbook (ie, by clicking a link, etc), the socket wouldn't disconnect and reconnect, so messages related to a new srcbook would be going to a misconfigured socket, leading to the dev server crashing.

This I think needs a fair amount of testing! As far as I can tell it works locally for me, but I am not sure if I am exercising all the cases.

packages/web/src/components/session-menu/index.tsx
```diff
@@ -0,0 +1,393 @@
...
+    contents: (props: SessionMenuPanelContentsProps) => <SessionMenuPanelSettings {...props} />,
+  },
+  // NOTE: re-enable this in the follow up change!
+  // {
+  //   name: 'secrets' as const,
+  //   icon: KeySquareIcon,
+  //   openWidthInPx: 480,
+  //   contents: () => <SessionMenuPanelSecrets />,
+  // },
+];
```
1egoman, line 51: This data structure controls rendering all the side panels. For now, the secrets panel is disabled.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,308 @@
...
+    <>
+      {/* FIXME: After closing this modal, `<body>` has `user-select: none;` on it? */}
+      <GenerateSrcbookModal
+        open={showGenSrcbookModal}
+        setOpen={setShowGenSrcbookModal}
+        openSrcbook={openSrcbook}
+      />
+      <ImportSrcbookModal open={showImportSrcbookModal} onOpenChange={setShowImportSrcbookModal} />
+      <DeleteSrcbookModal open={showDelete} onOpenChange={setShowDelete} session={props.session} />
+      <ExportSrcbookModal open={showSave} onOpenChange={setShowSave} session={props.session} />
```
1egoman, line 91: I have found that when any of these modals open and then close, the body has an inline `pointer-events: none;` style on it, and the app becomes somewhat non interactive:


https://github.com/user-attachments/assets/e083a0dd-0505-49ce-b9f9-b4963f1475ac

I need to look into this again and see what's going on. This doesn't happen on `main` so I think I introduced this inadvertently as part of my change.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,308 @@
...
+                      </DropdownMenuItem>
+                    );
+                  })}
+
+                {/* FIXME: how should more than 6 entries be rendered? */}
+                {props.srcbooks.length > 6 ? (
+                  <DropdownMenuItem asChild>
+                    <Link to="/">More...</Link>
+                  </DropdownMenuItem>
+                ) : null}
```
1egoman, line 144: Right now, only the 6 most recent entries are rendered in the inline srcbook picker dropdown. It's unclear from the designs if this is the intention or not, but given this probably should be length limited in some capacity to keep it from overflowing the screen, I limited it to 6, and if there's more than 6, rendered a `More...` item that just goes back to the home view.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,308 @@
...
+            <Button variant="icon" size="icon" className="active:translate-y-0">
+              <SocialDiscordIcon />
+            </Button>
+          </a>
+          <a href="https://github.com/srcbookdev/srcbook" target="_blank">
+            <Button variant="icon" size="icon" className="active:translate-y-0">
+              <SocialGithubIcon />
+            </Button>
+          </a>
+        </div>
```
1egoman, line 304: I think I saw in the mockups that the `SocialXIcon` entry should no longer be here? If I got that right I'll remove it.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,308 @@
...
+    const { result } = await createSrcbook({
+      path: props.baseDir,
+      name: 'Untitled',
+
+      // FIXME: copy the same language as the currently visible srcbook - should this be less
+      // implicit?
+      language: props.session.language,
+    });
+    openSrcbook(result.path);
+  }
```
1egoman, line 79: Note that for now, clicking "create srcbook" inline on the navbar will create a new srcbook with the same language as the current srcbook. I didn't see an explicit way this was handled in the designs - if there's a better way to do this that comes to mind, let me know.

benjreinhart: This is a good call out. I think we'll eventually probably have a modal or a blank page with a modal that asks you to select a few things (e.g., language, react or not if we choose to make that something user selects up front, possibly srcbook name, etc).

packages/web/src/components/navbar.tsx
1egoman: Figured this out, fixed in https://github.com/srcbookdev/srcbook/pull/267/commits/835e7f4128b9de34e6a91dec004cddff40aee776

packages/web/src/components/session-menu/table-of-contents-panel.tsx
```diff
@@ -0,0 +1,68 @@
...
+        <span>Table of contents</span>
+      </div>
+
+      <div className="max-w-60 text-tertiary-foreground pr-10">
+        {cells.map((cell) => {
+          const isRunningCell = cell.type === 'code' && cell.status === 'running';
+          return (
+            <div
+              key={cell.id}
+              className={cn(
```
benjreinhart, line 27: The font looks bigger here than the designs. I believe figma has it at text-sm (14px) but I think it's inheriting text-base (16px)

benjreinhart (commented):
Ooo boy this is looking 🔥 I left some comments but this is mostly ready to go!

One other thing: I don't see the icons at the bottom of the left navigation tab (the keyboard shortcuts and feedback).

Screenshot 2024-09-10 at 12 18 45 PM

packages/web/src/components/navbar.tsx
benjreinhart: Sounds good to me

packages/web/src/components/session-menu/secrets-panel.tsx
```diff
@@ -0,0 +1,12 @@
+// import { SessionMenuPanelContentsProps } from '.';
```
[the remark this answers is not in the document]
1egoman: Sure, I can remove this.

packages/web/src/routes/session.tsx
```diff
@@ -250,74 +295,86 @@ function Session(props: { session: SessionType; channel: SessionChannel; config:
...
+          openDepsInstallModal={() => setDepsInstallModalOpen(true)}
+          channel={channel}
+        />
 
-        {cells.map((cell, idx) => (
-          <div key={cell.id}>
+        {/* At the xl breakpoint, the sessionMenu appears inline so we pad left to balance*/}
+        <div className="grow shrink lg:px-0 pb-28">
+          <div className="max-w-[800px] mx-auto my-12 px-[32px]">
+            <TitleCell cell={titleCell} updateCellOnServer={updateCellOnServer} />
```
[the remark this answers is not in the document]
1egoman: It looks like this behavior currently happens in `main`:
Screenshot 2024-09-10 at 3 43 21 PM

That being said, I can take a look if desired - `word-wrap: break-word` might do what you are looking for here?

packages/web/src/components/navbar.tsx
1egoman: Removed.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+  return (
+    <>
+      {/* FIXME: After closing this modal, `<body>` has `user-select: none;` on it? */}
```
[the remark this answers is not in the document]
1egoman: Nope, forgot to remove this, nice catch!

benjreinhart (commented):
We'll need to create a changeset for this as well. Run `pnpm changeset` and follow the prompts. Let me know if that's confusing and I can help

https://github.com/srcbookdev/srcbook/blob/main/CONTRIBUTING.md#making-a-changeset

packages/web/src/routes/session.tsx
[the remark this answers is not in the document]
benjreinhart: Ok we can do that in separate PR then.

packages/web/src/components/navbar.tsx
```diff
@@ -0,0 +1,330 @@
...
+      <GenerateSrcbookModal
```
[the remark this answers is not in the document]
1egoman: I think what was going on here was that navigating to the new srcbook was going on in the background with the generate modal still open on top. Assuming that was the failure you were seeing, I think https://github.com/srcbookdev/srcbook/pull/267/commits/c7196596e410d1056c7f9d15892ab0981dc2f5a0 should fix it.

benjreinhart (approved):
:shipit:

packages/web/src/components/navbar.tsx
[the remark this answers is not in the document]
1egoman: Update: yep that was it!

PR #268, opened 2024-09-10
Use HOME_DIR constant to figure out database path

### LINKED ISSUE #245: Doesn't work on windows
I have tried both `npx srcbook start` and installing globally with npm, and neither works.  I get different errors depending on whether I attempt running the srcbook server in powershell vs. something like git bash.

Is this expected?

WesBosman:
### Issue
The `process.env.HOME` variable was undefined when running `npm install -g srcbook && srcbook start` on Windows. 
This was preventing the database from being created on Windows which also prevented the user from being able to launch srcbook.

### Solution
Use the `HOME_DIR` constant

Fixes #245  

Manually tested on mac and windows and the `HOME_DIR` constant seems to work on both.


benjreinhart (approved):
Looks perfect, thank you!

PR #274, opened 2024-09-12
move docs to focus on npx srcbook@latest instead of global install

versecafe:


benjreinhart (approved):
We'll also want to make these changes to `srcbook/README.md` as that is what users will see on npm.

PR #275, opened 2024-09-12
Small ux tweaks for prettier formatting

benjreinhart:
- [x] We had talked offline about using prettier logo since paint brush was non standard and didn't seem immediately obvious
- [x] Removed spinner because A) it was outside of button and so styles shifted when formatting and B) there is already a loading affect (code overlay) so I don't think both are critical
- [x] Small copy tweak


nichochar:
I think prettier is less obvious than paintbrush, but I realized that it's a good option since we don't have "any" formatter, that the user can configure, but specifically prettier. We can revisit the icon and affordance if we allow users to customize (like if we add biome support), which I don't think we need to think about anytime soon.

PR #277, opened 2024-09-12
support backend hmr

swk777:


packages/web/vite.config.ts
```diff
@@ -1,10 +1,29 @@
...
 import react from '@vitejs/plugin-react-swc';
+import chokidar from 'chokidar';
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [react()],
+  plugins: [
+    react(),
+    // reload on backend change
+    {
```
benjreinhart, line 11: Why do we need this on the client, to reload the websocket connection?

benjreinhart (commented):
Looks promising! Curious to get @versecafe's thoughts here

packages/configs/ts/base.json
```diff
@@ -9,7 +9,7 @@
     "isolatedModules": true,
     "useDefineForClassFields": true,
     "lib": ["es2022", "DOM", "DOM.Iterable"],
-    "module": "Preserve",
+    "module": "esnext",
```
benjreinhart, line 12: Can you explain this change? cc @versecafe

packages/web/vite.config.ts
versecafe: this should just be reloading the ws on the client whenever the API rebuilds in dev mode which is needed otherwise it's connected to a nonexistent session and breaks but @swk777 correct me if i'm wrong on that

versecafe: this should just be reloading the ws on the client whenever the API rebuilds in dev mode which is needed otherwise it's connected to a nonexistent session and breaks but @swk777 correct me if i'm wrong on that

packages/configs/ts/base.json
versecafe: @benjreinhart there shouldn't be a huge difference between esnext and preserve but removing preserve allows for some small transformations in tsc for cleaning things up, this should be a fine change and even make things easier but does mean a test to make sure the full bundled version is resolving correctly needs to be done to prevent a mishap like the 0.0.1 release bug we had

packages/web/vite.config.ts
benjreinhart: That's what I assume. I would be ok with this as a stop-gap but the better solution here is for the client side websocket client to re-establish connections when the connection is closed. Robust websocket clients use a combination of heartbeats and onclose listeners to attempt reconnect with a backoff strategy.

For now I'm ok to merge this but I would like to comment and/or file an issue to capture this.

packages/configs/ts/base.json
benjreinhart: Sweet, so long as it's all working with no major downsides LGTM

packages/web/vite.config.ts
benjreinhart: There is already some retry logic inside of the client, but I think this is retrying to send a message if it fails, so maybe it doesn't handle this case properly.

https://github.com/srcbookdev/srcbook/blob/main/packages/web/src/clients/websocket/client.ts#L10-L11

benjreinhart: https://github.com/srcbookdev/srcbook/issues/280

packages/configs/ts/base.json
swk777: Thank you @versecafe for the confirmation. I've discovered that "module: preserve" doesn't work well with vite-node watch mode. If necessary, we can consider keeping "module: esnext" specifically for the development mode @benjreinhart

packages/web/vite.config.ts
swk777: @versecafe is right, and as @benjreinhart mentioned, this approach is indeed better. I’ll work on implementing this later. 
Another point that deserves attention is that in the current design, the cellId is randomly generated, which changes after a reload. As a result, cell operations cannot be continued seamlessly across reloads.

benjreinhart: That's a good point, we need to fix those.

benjreinhart (approved):

PR #283, opened 2024-09-13
Goto Definition

versecafe:


packages/web/src/components/cells/code.tsx
```diff
@@ -693,6 +725,8 @@ function CodeEditor({
   readOnly: boolean;
 }) {
   const { theme, codeTheme } = useTheme();
+  const refs = useRef<ReactCodeMirrorRef>({});
```
benjreinhart, line 728: Is there a reason to use a react ref or react state here?

* The click handler takes the `view` as the second argument, so the ref is not needed?
* The browser `PointerEvent` passed to the `click` handler contains the necessary properties to determine if the `meta` or `altKey` is pressed.

If not necessary, than I think the code below works and is less complicated?

```typescript
Prec.high(
  EditorView.domEventHandlers({
    click: (e, view) => {
      // no need to run posAtCords calculations if user isn't cmd+clicking
      if (e.metaKey /* or e.altKey */) {
        const pos = view.posAtCoords({ x: e.clientX, y: e.clientY });
        if (pos) {
          view.focus();
          gotoDefinition(pos, cell, session, channel);
        }
      }
    },
  }),
),
```

One last thing... since the editor seems to insert a cursor at each place the user clicks while holding cmd, I would say that we either: Find a different way to handle multiple insertions (e.g., vscode uses `alt+click`) or we use `alt+click` instead of `meta+click` here? I'd prefer to emmulate vscode if possible since lots of people will know those bindings.

PR #284, opened 2024-09-13
Add ability to select session specific secrets

1egoman:
Adds a new "secrets" panel to the left side of the app where one can select which secrets apply to a given srcbook. By default, a srcbook has access to none of a user's defined secrets, and a user must manually enable each.

I've kept the implementation relatively straightforward for now - there's a new `secrets_to_sessions` table in sqlite that is used as a many-to-many bridge table between secrets and all sessions in a user's local `~/.srcbook/srcbooks` directory. I'm calling this out because in the future, some logic similar to [livebook stamping](https://news.livebook.dev/hubs-and-secret-management---launch-week-1---day-3-3tMaJ2xthttps://news.livebook.dev/hubs-and-secret-management---launch-week-1---day-3-3tMaJ2) might be a good idea to harden the security here, but for the time being this was deemed to be out of scope.

Screenshot 2024-09-13 at 2 29 30 PM

### DIFF
--- a/.changeset/moody-dragons-compare.md
+++ b/.changeset/moody-dragons-compare.md
@@ -2,4 +2,4 @@
 '@srcbook/web': patch
 ---
 
-added project specific secrets
+Secrets must now be enabled per Srcbook
--- a/.changeset/moody-dragons-compare.md
+++ b/.changeset/moody-dragons-compare.md
@@ -1,0 +1,5 @@
+---
+'@srcbook/web': patch
+---
+
+added project specific secrets
--- a/packages/api/config.mts
+++ b/packages/api/config.mts
@@ -46,32 +46,33 @@ export async function updateConfig(attrs: Partial<Config>) {
 
 export async function getSecrets(): Promise<Array<SecretWithAssociatedSessions>> {
   const secretsResult = await db.select().from(secrets);
-  const secretsToSessionResult = await db.select()
+  const secretsToSessionResult = await db
+    .select()
     .from(secretsToSession)
     .where(
       inArray(
         secretsToSession.secret_id,
-        secretsResult.map(secret => secret.id),
+        secretsResult.map((secret) => secret.id),
       ),
     );
 
-  return secretsResult.map(secret => ({
+  return secretsResult.map((secret) => ({
     name: secret.name,
     value: secret.value,
-    associatedWithSessionIds: (
-      secretsToSessionResult
-        .filter(secretToSession => secretToSession.secret_id === secret.id)
-        .map(secretToSession => secretToSession.session_id)
-    ),
+    associatedWithSessionIds: secretsToSessionResult
+      .filter((secretToSession) => secretToSession.secret_id === secret.id)
+      .map((secretToSession) => secretToSession.session_id),
   }));
 }
 
-export async function getSecretsAssociatedWithSession(sessionId: string): Promise<Record<string, string>> {
+export async function getSecretsAssociatedWithSession(
+  sessionId: string,
+): Promise<Record<string, string>> {
   const secretsResults = await getSecrets();
   return Object.fromEntries(
     secretsResults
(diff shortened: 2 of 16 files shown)


packages/api/config.mts
```diff
@@ -44,12 +44,35 @@ export async function updateConfig(attrs: Partial<Config>) {
...
+  return secretsResult.map(secret => ({
+    name: secret.name,
+    value: secret.value,
+    associatedWithSessionIds: (
+      secretsToSessionResult
+        .filter(secretToSession => secretToSession.secret_id === secret.id)
+        .map(secretToSession => secretToSession.session_id)
+    ),
+  }));
+}
```
1egoman, line 67: I opted to return a list of associated sessions in the root `/secrets` api response and use this same request to power the new secrets panel, which required some reshuffling of the response format.

packages/web/src/components/session-menu/secrets-panel.tsx
```diff
@@ -0,0 +1,139 @@
...
+        setSecretsList({ status: 'error' });
+        return;
+      }
+
+      setSecretsList({ status: 'complete', data: result });
+    };
+
+    // FIXME: it is possible for this to run twice in parallel?
+    run();
+  }, []);
```
1egoman, line 37: I couldn't find an example of data loading happening in anywhere in the application outside ofa react-router `loader` function. Given that this secrets list data is really only used in the secrets list panel, it seemed like the logical place to manage it, but in order to kick off the data fetching from here I had to add some relatively complex state management.

I can do a few things potentially here:
1. Leave it as it is (fwiw, I tend to be a fan of what I did here for its explicitness, but I recognize I'm probably in the minority on this view 😄 )
2. Switch this over to some other pre-built componet level data fetching tool like swr or react-query
3. Move this into the session page `loader` function and drill the state all the way down to this component, plus a way to "expire" it or oprimisitcally update it when changes are made to the switches

.changeset/moody-dragons-compare.md
```diff
@@ -0,0 +1,5 @@
+---
+'@srcbook/web': patch
+---
+
+added project specific secrets
```
benjreinhart, line 5: Nano nit but can we refine this a bit (will be in changelog). One possibility:

> Secrets must now be enabled per Srcbook

packages/api/drizzle/0008_add_secrets.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE `secrets_to_sessions` (
+	`id` integer PRIMARY KEY NOT NULL,
+	`session_id` text NOT NULL,
+	`secret_id` integer NOT NULL,
+	FOREIGN KEY (`secret_id`) REFERENCES `secrets`(`id`) ON UPDATE no action ON DELETE no action
```
benjreinhart, line 5: This needs a on delete cascade otherwise we cannot delete the secret on the secrets page without an exception (which currently crashes the server)

benjreinhart (commented):
As a first pass at this, looks good after you address the FK delete issue!

packages/api/drizzle/0008_add_secrets.sql
1egoman: Good catch! Should be fixed.

1egoman:
@benjreinhart I think I've addressed both of your comments!

packages/web/src/components/session-menu/secrets-panel.tsx
benjreinhart: No this is great for now. We have held off on fancier data fetching / state management solutions until we:

1. Have lots of examples in our code base to best inform us how we'd like to handle this
2. Have any sort of perf considerations (right now, we only run locally, so caching or anything clever isn't that helpful)

:shipit:

benjreinhart (approved):
Awesome, thank you 🙏🏻

PR #286, opened 2024-09-13
Tweak theme to match ours for mermaid diagrams

benjreinhart:
I only tested on two diagram types. There are a number of other settings and colors for different things, so I'm sure this is far from perfect. But it's mostly there with minimal effort for now.

Screenshot 2024-09-13 at 1 01 50 PM

Screenshot 2024-09-13 at 1 01 56 PM


versecafe (approved):
Looks much cleaner

PR #303, opened 2024-09-16
Improve mermaid rendering

benjreinhart:
I was noticing some wonky behavior with Mermaid rendering. I would occasionally see one or both of the following:

* With multiple diagrams, sometimes the renderer would render all of them in the same location, even if they were supposed to be rendered at different points in the markdown
* Sometimes (with multiple diagrams especially) I would see one or two diagrams render but not the others

Then I noticed the keys were fixed even though there could be multiple rendering in the same markdown block. We were getting unique key warnings in the console.

### DIFF
--- a/.changeset/brown-peas-hope.md
+++ b/.changeset/brown-peas-hope.md
@@ -1,0 +1,5 @@
+---
+'@srcbook/web': patch
+---
+
+Improve mermaid rendering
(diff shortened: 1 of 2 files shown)


versecafe (approved):
might want to shorten the keys they'll end up huge but it'll work for now and shouldn't rerender a ton so not a huge issue

PR #306, opened 2024-09-17
Click to Go to Def

versecafe:


versecafe:
- Does not support recursive def view,
- TSServer not running over the content in external view
- In larger files it's a bit of a pain since it just opens the file itself

pnpm-lock.yaml
```diff
@@ -6,18 +6,9 @@ settings:
 
 catalogs:
   default:
-    '@ai-sdk/anthropic':
```
benjreinhart, line 9: I don't see any package.json changes, why the change here?

benjreinhart (commented):
I'm currently unable to get this to work on my machine, maybe because something with my environment is screwed up? I also cannot seem to `pnpm build --force`.

packages/api/server/http.mts
```diff
@@ -58,6 +58,32 @@ router.post('/disk', cors(), async (req, res) => {
...
+router.post(
+  '/file',
+  cors({
+    origin: ['http://localhost:5173', 'http://localhost:5713'],
+    methods: ['GET', 'POST', 'PUT', 'DELETE'],
+    allowedHeaders: ['Content-Type', 'Authorization'],
+    credentials: true,
+  }),
+  async (req, res) => {
+    console.log('request hit');
```
benjreinhart, line 70: Looks like this can be removed

packages/web/src/components/cells/code.tsx
```diff
@@ -675,6 +677,47 @@ function tsLinter(
   });
 }
 
+function gotoDefinition(
```
benjreinhart, line 680: There are multiple console.logs in here we can remove before merge

packages/api/server/http.mts
```diff
@@ -58,6 +58,32 @@ router.post('/disk', cors(), async (req, res) => {
   }
 });
 
+router.post(
+  '/file',
+  cors({
```
benjreinhart, line 63: Can we use cors here the way we used it for all the other functions? This doesn't work for me:

Screenshot 2024-09-17 at 4 18 32 PM

pnpm-lock.yaml
versecafe: lockfile rebuild maybe not sure why there are changes there

benjreinhart: Can we remove htem?

versecafe: @benjreinhart looking at it that's some weird bugged state, fixed it now

packages/web/src/components/cells/code.tsx
```diff
@@ -675,6 +677,44 @@ function tsLinter(
...
+  editor: (content: string) => void,
+) {
+  async function gotoDefCallback({ response }: TsServerDefinitionLocationResponsePayloadType) {
+    channel.off('tsserver:cell:definition_location:response', gotoDefCallback);
+    if (response === null || response === undefined || Object.keys(response).length === 0) {
+      console.log('no response');
+      return;
+    }
+    const local: boolean =
+      response.file.includes('.srcbook/srcbooks') && !response.file.includes('node_modules');
```
benjreinhart, line 694: I assume these lines will not work on windows given the path separator being hardcoded? Can we have the server pass something down that is easier for the client to work with?

If the `TsServerDefinitionLocationResponsePayloadType` had a `cellId?: string` property on it, then we could simply do:

```typescript
async function gotoDefCallback(location: TsServerDefinitionLocationResponsePayloadType) {
  // ...
  if (location.cellId) {
    document.getElementById(`cell-${location.cellId}`)?.scrollIntoView({ behavior: 'smooth' });
  }
  //...
}
```

If `cellId` is not a string, then we know it needs a file look up.

packages/web/src/components/cells/code.tsx
```diff
@@ -711,6 +755,29 @@ function CodeEditor({
...
+        },
+        keyup: (e) => {
+          if (e.key === 'Alt') {
+            setCmdKey(false);
+          }
+        },
+        click: (e, view) => {
+          const pos = view.posAtCoords({ x: e.clientX, y: e.clientY });
+          if (pos && cmdKey) {
+            extensions.push(EditorView.editable.of(false));
```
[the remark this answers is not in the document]
benjreinhart: I still don't know why we need these?

packages/web/src/components/cells/code.tsx
```diff
@@ -257,7 +259,7 @@ export default function CodeCell(props: {
           ) : (
             <ResizablePanelGroup direction="vertical">
               <ResizablePanel style={{ overflow: 'scroll' }} defaultSize={60}>
-                <div className={cn(cellMode !== 'off' && 'opacity-50')}>
+                <div className={cn(cellMode !== 'off' && 'opacity-50')} id={cell.filename}>
```
benjreinhart, line 262: Are there characters supported in filenames that would not be supported by html ids (at least without escaping)?

Can we use the cell id instead? Right now it changes between server restarts but we definitely need to fix that.

packages/web/src/components/cells/code.tsx
```diff
@@ -729,17 +782,40 @@ function CodeEditor({
...
+      <CodeMirror
+        value={cell.source}
+        theme={codeTheme}
+        extensions={extensions}
+        onChange={(source) => {
+          updateCellOnClient({ ...cell, source });
+          updateCellOnServerDebounced(cell, { source });
+        }}
+      />
+      <Dialog open={isModalOpen} onOpenChange={setIsModalOpen}>
```
benjreinhart, line 804: Fine to merge with this but in general it is probably cleaner to have one dialog rendered for this purpose rather than N.

packages/web/src/components/cells/code.tsx
benjreinhart: The server and client are communicating over a session, so they both always know what the current session is.

packages/api/server/ws.mts
```diff
@@ -788,11 +788,15 @@ async function getDefinitionLocation(payload: TsServerDefinitionLocationRequestP
...
 
-  wss.broadcast(`session:${session.id}`, 'tsserver:cell:definition_location:response', {
-    response: {
-      ...body[0],
-    },
-  });
+  const res = {
+    response: body[0]
+      ? {
+          ...body[0],
```
benjreinhart, line 794: Why do we need to create a copy of this object?

benjreinhart (approved):

packages/api/server/ws.mts
versecafe: needs to be typed as `{respinse: null}` or `{response: { ...body[0] }}` is there another way to do that which avoids a copy?

benjreinhart: Does `body[0]` work?

```typescript
const res = {
  response: body[0] ? body[0] : null, // or body[0] || null
};
```

versecafe: yep lmao this was something stupid I did when fighting zod very confused over something it was doing cleaned up

PR #307, opened 2024-09-17
Import / export modal improvements

1egoman:
Adds the ability within the import and export modals to import and export via ways other than saving to the filesystem. In addition, removes non-native file components and ports to native browser file handling logic wherever possible.

## Import modal
The import modal now contains three tabs, allowing a user to import via drag+drop or clicking to select a file, entering a url (which the backend will make a `GET` request to), or copying and pasting a `*.src.md` file into a textarea.

Screenshot 2024-09-17 at 3 36 57 PM
Screenshot 2024-09-17 at 3 37 01 PM
Screenshot 2024-09-17 at 3 37 06 PM

## Export modal
A user can export a srcbook as two things: either a file (which when the `Download File` button is clicked, downloads the file to their downloads directory), or plain text (which can be copied to the user's clipboard).
Screenshot 2024-09-17 at 3 54 23 PM
Screenshot 2024-09-17 at 3 54 28 PM


packages/api/server/http.mts
```diff
@@ -41,23 +41,6 @@ const router = express.Router();
...
-    dirname = dirname || config.baseDir;
-    const entries = await disk(dirname, '.src.md');
-    return res.json({ error: false, result: { dirname, entries } });
-  } catch (e) {
-    const error = e as unknown as Error;
-    console.error(error);
-    return res.json({ error: true, result: error.stack });
-  }
-});
-
```
1egoman, line 60: I think I've removed all the disk / custom file picker implementation stuff. If you see anything else I've missed, please call it out.

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
+            <button
+              className="flex flex-col gap-4 w-full items-center justify-center h-[160px] border border-dashed rounded-md cursor-default"
+              onClick={onDownloadSrcbook}
+            >
+              <div className="flex flex-col items-center gap-2 text-tertiary-foreground">
+                <NotebookIcon size={24} />
+                <code className="text-xs">{downloadFileName}</code>
+              </div>
+              <Button>Download File</Button>
+            </button>
```
1egoman, line 412: Note that there is a slight bit of functionality being lost here - previously, users could save a `*.src.md` file anywhere on their filesystem. Now, when it is downloaded, it only ends up in their downloads directory. Unless I'm missing something (and without reaching for [chrome specific apis](https://developer.chrome.com/docs/capabilities/web-apis/file-system-access)) I'm not sure there's a way around this.

benjreinhart: I'm not an expert with these newer APIs, but the [showSaveFilePicker](https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker) function seems to be what we want. It may not work in all browsers, but if it works in chrome and edge that should solve for at least 80%, if not 95%+.

Perhaps we can check for its existence and use it, otherwise fallback to what we have in this PR?

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
         </DialogHeader>
-        <ExportLocationPicker onSave={onSave} title={getTitleForSession(session)} />
-        {error && <p className="text-destructive-foreground">{error}</p>}
+
+        <Tabs value={activeTab} onValueChange={(tab) => setActiveTab(tab as 'file' | 'text')}>
+          <div className="border-b mb-4">
+            <TabsList>
+              <TabsTrigger value="file">
+                <div className="flex items-center gap-2">
+                  <FilesIcon size={18} />
```
benjreinhart, line 389: Can we use the "file-down" icon here?

packages/web/src/routes/settings.tsx
```diff
@@ -19,23 +16,7 @@ import { Switch } from '@/components/ui/switch';
...
 
-async function loader() {
-  const { result: diskResult } = await disk({});
-
-  return {
-    ...diskResult,
-  };
-}
-
-async function action({ request }: { request: Request }) {
```
benjreinhart, line 30: This makes me so happy

packages/api/server/http.mts
```diff
@@ -104,17 +87,24 @@ router.delete('/srcbooks/:id', cors(), async (req, res) => {
 // Import a srcbook from a .src.md file or srcmd text.
 router.options('/import', cors());
 router.post('/import', cors(), async (req, res) => {
-  const { path, text } = req.body;
+  const { path, text, url } = req.body;
 
   if (typeof path === 'string' && !isSrcmdPath(path)) {
     return res.json({ error: true, result: 'Importing only works with .src.md files' });
   }
+  if (typeof url === 'string' && !isSrcmdPath(url)) {
```
benjreinhart, line 95: I think it's more robust to make no assumptions about paths. For now the hub will have this format, but that may not always be the case. More importantly, in theory, these could be hosted or available anywhere by anyone (even though today that's very unlikely)

packages/api/server/http.mts
```diff
@@ -104,17 +87,24 @@ router.delete('/srcbooks/:id', cors(), async (req, res) => {
...
+  }
 
   try {
     if (typeof path === 'string') {
       posthog.capture({ event: 'user imported srcbook from file' });
       const srcbookDir = await importSrcbookFromSrcmdFile(path);
       return res.json({ error: false, result: { dir: srcbookDir } });
+    } else if (typeof url === 'string') {
+      posthog.capture({ event: 'user imported srcbook from url' });
+      const srcbookDir = await importSrcbookFromSrcmdUrl(url);
```
benjreinhart, line 106: yessssss. Do you mind making a follow up note once this PR goes through to update [the CLI code](https://github.com/srcbookdev/srcbook/blob/main/srcbook/src/cli.mts#L81-L131) to use this?

packages/api/server/http.mts
benjreinhart: The thing that matters is that we fail if we cannot successfully parse it, which I think is still happening?

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
+          <TabsContent className="mt-0" value="file">
+            <button
+              className="flex flex-col gap-4 w-full items-center justify-center h-[160px] border border-dashed rounded-md cursor-default"
+              onClick={onDownloadSrcbook}
+            >
+              <div className="flex flex-col items-center gap-2 text-tertiary-foreground">
+                <NotebookIcon size={24} />
+                <code className="text-xs">{downloadFileName}</code>
+              </div>
+              <Button>Download File</Button>
```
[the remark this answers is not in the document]
versecafe: for the outer layer you can and probably should use a div with aria attributes to mark it as button like and do click detection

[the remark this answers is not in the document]
1egoman: Oops! I'm just going to get rid of this nested click handler situation here, upon reflection I'm not sure that it makes the user experience much better if at all.

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -20,37 +25,148 @@ export function ImportSrcbookModal({
...
+      parsedUrl = new URL(url);
+    } catch (err) {
+      console.error(`Cannot parse ${url} as url:`, err);
+      setLoading(false);
+      setError(`Cannot parse ${url} as a url!`);
+      return;
+    }
+
+    const fileName = (
+      parsedUrl.pathname.length > 0 ? parsedUrl.pathname.split('/').at(-1)! : url
```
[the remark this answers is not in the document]
1egoman: Ok, sounds good. For what it's worth, if the validation you are referring to is the name length check right around like 119, this was pre-existing behavior, but I'll see if I can move this check serverside. I agree having all the validation in one place is probably a good idea.

packages/web/src/components/import-export-srcbook-modal.tsx
benjreinhart: @versecafe is there a reason we can't we use it for browsers that support it and then fallback to what is here for browsers that don't?

The vast majority of people use chrome, so as the comment above states I would expect this to cover almost everyone. I'm not sure what ladybird is but a google search says "It is planning to have an alpha release in 2026" so I'm not too concerned about this affecting anyone

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -65,12 +181,94 @@ export function ImportSrcbookModal({
...
+            </div>
+          </TabsContent>
+          <TabsContent className="mt-0" value="clipboard">
+            <div className="flex flex-col gap-4">
+              <Textarea
+                ref={clipboardTextareaRef}
+                value={clipboard}
+                onChange={(event) => setClipboard(event.target.value)}
+                placeholder="Paste clipboard here"
+                className="h-[112px] resize-none"
```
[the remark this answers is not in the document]
1egoman: Ok, sounds good!

(as a counterargument to consider, the reason I didn't make this one resizable was because it seemed like it would be rare for a user to want to scroll through the raw text, and would likely only look at a preview and copy it to the clipboard)

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
+              </div>
+              <Button>Download File</Button>
+            </button>
+          </TabsContent>
+          <TabsContent className="mt-0" value="text">
+            <div className="flex flex-col gap-4">
+              <Textarea
+                ref={clipboardTextareaRef}
+                value={srcbookText.status === 'complete' ? srcbookText.text : ''}
+                className="font-mono text-xs whitespace-pre h-[112px] resize-none"
```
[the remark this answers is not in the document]
1egoman: Sounds good.

(FWIW, I had figured that wouldn't be that big of a deal as a user would just be copying and pasting out of that box, and wouldn't be deeply analyzing the text contents)

packages/web/src/components/import-export-srcbook-modal.tsx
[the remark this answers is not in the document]
benjreinhart: Yeah it's not a huge thing, most people probably won't. Maybe since I'm a developer of this tool and want to confirm things look right, I'm personally more likely to glance through it

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
+  // When hiding the modal, reset all controls
+  useEffect(() => {
+    if (!open) {
+      return;
+    }
+
+    setActiveTab('file');
+  }, [open]);
+
+  const downloadFileName = useMemo(() => {
```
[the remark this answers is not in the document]
1egoman: Ah, I must have missed this. Thanks for calling it out, and I agree that makes a lot more sense.

packages/web/src/components/import-export-srcbook-modal.tsx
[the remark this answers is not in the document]
benjreinhart: Yeah may well have been happening before but given we're touching this code I'd say let's replace with server-side + zod validation if quick to do.

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -65,12 +181,94 @@ export function ImportSrcbookModal({
...
+        <Tabs
+          value={activeTab}
+          onValueChange={(tab) => setActiveTab(tab as 'file' | 'url' | 'clipboard')}
+        >
+          <div className="border-b mb-4">
+            <TabsList>
+              <TabsTrigger value="file">
+                <div className="flex items-center gap-2">
+                  <FilesIcon size={18} />
+                  From Filesystem
```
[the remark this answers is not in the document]
1egoman: No preference, sounds good!

packages/api/server/http.mts
1egoman: Ok, sounds good!

> The thing that matters is that we fail if we cannot successfully parse it, which I think is still happening?
Yep, that is still happening!

packages/api/server/http.mts
1egoman: Unless you object, I'll just include that change in this pull request - it doesn't seem that complex, and I'd rather just get it done.

benjreinhart: Sounds good

1egoman: FWIW, I think this change should be here: https://github.com/srcbookdev/srcbook/pull/307/commits/4892f66048957558c644e37d4851a963e7e982f1

packages/web/src/components/import-export-srcbook-modal.tsx
1egoman: To close the loop on this one, here's what I did to use the web filesystem api while falling back to the old download behavior if this doesn't exist: https://github.com/srcbookdev/srcbook/pull/307/commits/ec54805f0c7f771c8990d39695b08b0ecaa37dd0

benjreinhart (approved):

packages/web/src/components/import-export-srcbook-modal.tsx
```diff
@@ -86,33 +284,149 @@ export function ExportSrcbookModal({
...
+  function onDownloadSrcbook() {
+    if (srcbookText.status !== 'complete') {
+      return;
+    }
+
+    const blob = new Blob([srcbookText.text]);
+
+    const url = URL.createObjectURL(blob);
+    const link = document.createElement('a');
+    link.setAttribute('href', url);
```
nichochar, line 373: Damn this is weird, but it works!

nichochar:
Tried this out, it works great.

A minor design nit (not worth fixing at this point) is the vertical resizing when going from "import URL" tab to any of the others. We could fix the height of the entire modal to be the taller version at all times as a cheap fix.
Thanks @1egoman !

Downloading mechanism worked great on my mac with Arc (chromium)

PR #310, opened 2024-09-19
Add ability to create secrets within a Srcbook

benjreinhart:
We want to make it quick and easy to add a new secret on the Srcbook page.

### Notes

* Since this is all running locally right now and thus response times are < 10ms, I simplified the existing code to ignore the loading states which made it easier to add the other logic (see first commit in this PR).
* This isn't perfect, the form is minimal. There's no link to the secrets page, which has more functionality. The UI is probably not great. We can work with our designer to improve this.

CleanShot 2024-09-19 at 00 04 26

### DIFF
--- a/packages/web/src/components/session-menu/secrets-panel.tsx
+++ b/packages/web/src/components/session-menu/secrets-panel.tsx
@@ -1,8 +1,5 @@
-// import { SessionMenuPanelContentsProps } from '.';
-
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 import { toast } from 'sonner';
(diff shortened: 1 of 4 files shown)


nichochar:
I wonder if we should soon consider using some of those libraries that help with loading state (react-query, tanstack-query, etc..).

versecafe (approved):

1egoman (approved):

1egoman:
> I wonder if we should soon consider using some of those libraries that help with loading state (react-query, tanstack-query, etc..).

I would put my vote in for this, if only so that the requisite intermediate states are captured for the future with minimal amounts of boilerplate. No strong preference on the tool!

PR #313, opened 2024-09-19
[FIX]:299 - uncaught exception session with id 'Not Found' after dele…

BeRecursive22:
…ting an srcbook

fix: Handle session not found errors in TsServer event handlers

- Added try-catch blocks in onSemanticDiag and onSuggestionDiag handlers
- Gracefully handle and log errors when findSession fails
- Prevent server crashes due to unhandled session not found errors


.changeset/six-vans-study.md
```diff
@@ -0,0 +1,5 @@
+---
+'@srcbook/api': patch
+---
+
+fix(api): Handle session cleanup after srcbook deletion. Resolved an issue where deleting an srcbook would lead to an uncaught exception due to an orphaned session. The fix ensures proper cleanup of associated sessions when an srcbook is deleted, preventing the "Session with id not found" error.
```
benjreinhart, line 5: Can we simplify this message to something like "Fix uncaught exceptions that can occur after Srcbook deletion" ?

benjreinhart (approved):

.changeset/six-vans-study.md
benjreinhart: This description is better placed in the PR description, with a short message that goes in the Changelog (which will link back to the PR)

BeRecursive22: @benjreinhart sure. But it seems like the issue is much bigger than what I anticipated previously. It looks like a race condition going on between tsserver and actual session deletion which is leading to this error

BeRecursive22: @benjreinhart this simply order change of session deletion won't make it go away. 

The issue, I believe is, not properly handling the errors for onSemanticDiag and onSuggestionDiag event when the session id is not found. 

image

At least, the server didn't crash when I added try...catch to catch the errors.

packages/api/server/ws.mts
```diff
@@ -124,6 +124,7 @@ async function cellExec(payload: CellExecPayloadType) {
   const session = await findSession(payload.sessionId);
   const cell = findCell(session, payload.cellId);
 
+  console.log('inside cellExec');
```
nichochar, line 127: Can we remove this?

nichochar:
Approved, this is looking good to go!

PR #318, opened 2024-09-20
feat: add go to definiton keyboard shortcuts entry

1egoman:
Note that according to vscode's documentation, it is "Go to definition" and not "Goto Definition", so I went with that: https://code.visualstudio.com/Docs/editor/editingevolved#_go-to-definition

Screenshot 2024-09-20 at 4 01 47 PM

### DIFF
--- a/packages/web/src/components/keyboard-shortcuts-dialog.tsx
+++ b/packages/web/src/components/keyboard-shortcuts-dialog.tsx
@@ -45,6 +45,7 @@ export default function KeyboardShortcutsDialog({
               <ShortcutRow keys={['alt', '↑']} description="move lines up" />
               <ShortcutRow keys={['alt', '↓']} description="move lines down" />
               <ShortcutRow keys={['shift', 'alt', 'f']} description="format code using Prettier" />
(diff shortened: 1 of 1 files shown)


packages/web/src/components/keyboard-shortcuts-dialog.tsx
```diff
@@ -45,6 +45,7 @@ export default function KeyboardShortcutsDialog({
               <ShortcutRow keys={['alt', '↑']} description="move lines up" />
               <ShortcutRow keys={['alt', '↓']} description="move lines down" />
               <ShortcutRow keys={['shift', 'alt', 'f']} description="format code using Prettier" />
+              <ShortcutRow keys={['alt', 'click']} description="Go to definition" />
```
benjreinhart, line 48: nano nit but the other descriptions all use lower case. Don't care either way but would prefer consistency

benjreinhart (approved):

packages/web/src/components/keyboard-shortcuts-dialog.tsx
1egoman: Oops, fixed

PR #319, opened 2024-09-20
Editor Completions

versecafe:


packages/web/src/components/cells/get-completions.ts
```diff
@@ -0,0 +1,75 @@
...
+//     ],
+//   };
+// }
+
+import { SessionChannel } from '@/clients/websocket';
+import { CompletionContext, CompletionResult } from '@codemirror/autocomplete';
+import { CodeCellType, TsServerCompletionEntriesPayloadType } from '@srcbook/shared';
+import { mapCMLocationToTsServer } from './util';
+
+export function myCompletions(
```
benjreinhart, line 28: Naming nit: maybe "getCompletions" ?

packages/web/src/components/cells/get-completions.ts
```diff
@@ -0,0 +1,75 @@
+// import { SessionChannel } from '@/clients/websocket';
```
benjreinhart, line 1: can we remove all these comments?

benjreinhart (approved):
Small amount of testing but it worked like a charm! So dope :shipit:

PR #324, opened 2024-09-25
Add new `@srcbook/components` package, and move the `TitleCell` inside

1egoman:
This change is a first pass at moving the `TitleCell` into a new `@srcbook/components` package, which is planned to be separately published and used in future srcbook hub interfaces.


packages/cells/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "prepublishOnly": "npm run build"
+  },
+  "dependencies": {
+    "clsx": "^2.1.1",
+    "tailwind-merge": "^2.5.2"
+  },
+  "peerDependencies": {
+    "react": "*",
+    "@types/react": "*"
+  }
```
1egoman, line 21: I did my best here based on what I've done in the past, but I'm not sure this is right.

1. In theory, the versions of `clsx` and `tailwind-merge` could be different than the versions in `@srcbook/web` - this might be a good property (not having to update everything in lock step means easier migrations), or it might be not (it could lead to duplicate code in the bundle). It also just might not matter right now too 😄 . If the same version is desired, it seems like that might be a place to use that pnpm catalogs feature?

2. In the past with dependency packages like this, I've set react as a peerdependency with little to no explicit version requirement to reduce dependency complexity / cognitive overhead. This is because In the future the hub project would need to be kept up to date and picking too strict of a version range will likely (I've had some experience here) make this difficult. That being said, I could see why such a wide open version range would not be ideal - so, if something more strict would be preferred, maybe the thing to do here is use that pnpm catalogs feature and define a global version of react that could be both used here and for the react version in the web package.

packages/cells/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "outDir": "dist",
+    "jsx": "react-jsx",
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["src/*"]
+    }
+  },
+  "include": ["index.tsx", "src/**/*.{ts.tsx}"],
+  "exclude": ["node_modules", "dist"]
+}
```
1egoman, line 15: The typescript configuration ended up being a hybrid between the `@srcbook/shared` file and the `@srcbook/web` file. Given this is the first web-specific package, let me know if there's any tweaks that should be made here.

packages/web/tailwind.config.js
```diff
@@ -1,9 +1,14 @@
...
+const path = require('path');
 
 module.exports = {
   darkMode: ['class'],
-  content: ['./index.html', './src/**/*.{css,ts,tsx}'],
+  content: [
+    './index.html',
+    './src/**/*.{css,ts,tsx}',
+    path.join(path.dirname(require.resolve('@srcbook/cells')), '**/*.{css,js,ts,tsx}'),
+  ],
```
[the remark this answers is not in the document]
benjreinhart: I agree this doesn't seem great but i also don't really know what the alternatives are. 

cc @versecafe do you have any thoughts on how we should best handle this?

packages/cells/package.json
benjreinhart: Should we also make clsx and tailwind-merge peer deps? I realize that's maybe not ideal in some cases but given we aren't expecting this to be a very generic package (mostly just for our internal use) it seems reasonable?

In either case, I don't think we need to overthink these things too much, at least assuming we don't blow up our asset files with duplicate dependencies

cc @versecafe

benjreinhart (commented):
I don't see a reason why we need a `cells` specific package as opposed to a more generic `ui` or `components` package that we can just throw any reusable components in. Are there good reasons to have a separate one? If not maybe we just make one generic UI package for now?

1egoman:
> I don't see a reason why we need a cells specific package as opposed to a more generic ui or components package that we can just throw any reusable components in. Are there good reasons to have a separate one? If not maybe we just make one generic UI package for now?

No, that was just what came to mind to me initially after our discussion. I'll rename this existing package to something like `@srcbook/components` (that probably makes more sense as a name I think since cells aren't ui atoms) and put both cells and ui components in here.

benjreinhart (approved):

nichochar (approved):

packages/components/.eslintrc.cjs
```diff
@@ -0,0 +1,13 @@
+/** @type {import("eslint").Linter.Config} */
+module.exports = {
+  root: true,
+  extends: [require.resolve('@srcbook/configs/eslint/library.js')],
```
versecafe, line 4: should use `@srcbook/configs/eslint/react.js` otherwise the accessibility (a11y) rules and the react-hook rules will be missed and jest types may break

packages/components/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "scripts": {
+    "prebuild": "rm -rf ./dist",
+    "build": "tsc",
+    "dev": "tsc --watch --project .",
+    "lint": "eslint . --max-warnings 0",
+    "check-types": "tsc",
+    "prepublishOnly": "npm run build"
+  },
+  "dependencies": {
+    "clsx": "^2.1.1",
```
versecafe, line 15: Should these be peerDeps since they're an external thing, and actually should shared have a utils section for things like `cn()`

packages/components/src/ui/heading.tsx
```diff
@@ -45,7 +45,6 @@ export function EditableH1(props: {
   return (
     <div>
       <h1
-        // eslint-disable-next-line jsx-a11y/no-noninteractive-element-to-interactive-role -- messy fix should be reworked
```
versecafe, line 48: This is still needed, the only reason this wasn't throwing a warning was because the wrong eslint config was used

versecafe (changes requested):
@1egoman the peerDeps comment is more of a question but the eslint issue should be fixed

benjreinhart (approved):

PR #302, opened 2024-09-16
Revert "support backend hmr"

benjreinhart:
Reverts srcbookdev/srcbook#277

@swk777 after merge, i ran `pnpm dev` and it spins up the API server twice, the second one failing due to port in use. I'm going to revert until that's fixed. My apologies, I should have verified this before merging.

PR #326, opened 2024-09-29
support backend hmr

swk777:
In the previous PR #277 #302 , @benjreinhart discovered an issue where the port was in use. Although I couldn’t reproduce it locally, there are indeed one or two "api" rebuilds when running "pnpm dev". After reviewing the setup, I found it probably relates to the "shared" build process triggering api HMR because the api project references the shared project.  I initially attempted to use Turbo’s deps configuration to control the startup order, but since both projects are persistent, deps can’t be configured for them and adding some delay to api dev looks a bit wield.To resolve this, I used "shared" incremental builds. 
If there’s a better approach, please let me know.


packages/web/package.json
```diff
@@ -63,6 +63,7 @@
     "@types/react-dom": "^18.3.0",
     "@vitejs/plugin-react-swc": "^3.7.0",
     "autoprefixer": "^10.4.20",
+    "chokidar": "^3.6.0",
```
benjreinhart, line 66: They just released 4.0.0 of the lib and it looks to have substantial improvements. Can we use that version?

benjreinhart (commented):
Code lgtm, I will pull this down later to test.

@versecafe would be great to have you look at these changes as well.

packages/web/package.json
versecafe: With @benjreinhart on this, bumping to 4.0 gets rid of a bunch of node deps and makes things simpler, but their changelog does state glob support was removed so I'm not sure what would be needed to keep the `../api/**/*.mts` behaviour

swk777: I’ve upgraded chokidar to v4 and adapted the codebase @benjreinhart @versecafe

versecafe:
All seems to be working smoothly without any issues in testing

PR #329, opened 2024-10-01
Port code cell to components

1egoman:
Ports the `CodeCell` from `@srcbook/web` to `@srcbook/components`.

The code cell has a very tight dependency on the session websocket connection, so I went through a number of different approaches to try to find the best abstraction boundary. What I landed on was creating a wrapper called `ControlledCodeCell` in the `@srcbook/web` package which then renders the regular `CodeCell` located in `@srcbook/components`, which has been updated to be completely presentational, and take (quite) a list of props to control how it renders.

A few other notes:
- I haven't tested this as effectively as I would like. From a cursory look things seem to work in the editor still, however given the amount of code that has changed I almost certainly have missed something.
- I ended up moving the cell context into the `@srcbook/components` package.
  - `HeaderCell` / `MarkdownCell` don't use the cell context currently (and probably should be updated to use it to be consistient?), while `CodeCell` does.
  - I went back and forth here on where this should be and neither seemed clearly like a better option. (should the context live in `@srcbook/components` so it can be used to avoid prop drilling somewhat, or should it live in `@srcbook/web` to make the `@srcbook/components` code as minimally complex as possible?)
- The go to definition modal logic has been moved out of the CodeCell and into the `Session` route


packages/components/src/components/cell-output.tsx
```diff
@@ -1,11 +1,10 @@
...
 import { Ban, Maximize, Minimize, PanelBottomClose, PanelBottomOpen, Sparkles } from 'lucide-react';
 import { CodeCellType, PackageJsonCellType, TsServerDiagnosticType } from '@srcbook/shared';
-import { cn } from '@/lib/utils';
-import { Tabs, TabsList, TabsTrigger, TabsContent } from '@/components/ui/underline-flat-tabs';
-import { useCells } from '@/components/use-cell';
-import { OutputType, StdoutOutputType, StderrOutputType, CellModeType } from '@/types';
+import { cn } from '../lib/utils';
+import { Tabs, TabsList, TabsTrigger, TabsContent } from './ui/underline-flat-tabs';
+import { useCells } from './use-cell';
+import { OutputType, StdoutOutputType, StderrOutputType, CellModeType } from '../types';
```
1egoman, line 7: Note that the `@` prefix notation won't work right in child packages - that syntax is included in the bundled output and therefore is picked up by the package resolution logic in the parent package 🤦 . I couldn't figure out how to get this to work in `@srcbook/components` so I converted everything in that package over to use relative imports.

packages/components/src/components/cells/code.tsx
```diff
@@ -0,0 +1,636 @@
...
+  onGenerate: () => void;
+  onGetDefinitionContents: (pos: number, cell: CodeCellType) => Promise<string>;
+  onRevert: () => void;
+  onRunCell: () => void;
+  onStopCell: () => void;
+  onUpdateFileName: (filename: string) => void;
+  prompt: string;
+  showStdio: boolean;
+  updateCellOnServer: (cell: CodeCellType, attrs: CodeCellUpdateAttrsType) => void;
+  editorExtensions: Array<Extension>;
```
1egoman, line 78: Ben and I talked about this I beleive, but a big win in removing complexity was to move all the extension setup out of the code cell and up one level up, which means that all the tendrils that the editor specific overlays / autocomplete / etc have into the tsserver can be sidestepped in this abstraction. This worked out well!

packages/components/src/components/cells/code.tsx
```diff
@@ -0,0 +1,636 @@
...
+                  <CodeEditor cell={props.cell} extensions={[]} codeTheme={props.codeTheme} />
+                ) : (
+                  <CodeEditor
+                    // FIXME: make sure that the changes this should be causing in `props.editorExtensions` actually are working
+                    // readOnly={['generating', 'prompting', 'formatting'].includes(props.cellMode)}
+                    cell={props.cell}
+                    updateCellOnServer={props.updateCellOnServer}
+                    extensions={props.editorExtensions}
+                    codeTheme={props.codeTheme}
+                  />
```
1egoman, line 248: This is a place where some more testing could take place.

Because I made changes to where the extensions were set up (mentioned in an earlier comment), that had implications on which were enabled in certain modes and in which the editor was read only. I'm like 95% sure I did it right but having somebody with a strong grasp on exactly the expected behavior make sure the editor renders right in each `cellMode` would be helpful.

packages/components/src/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+export * from './components/ui/textarea';
+export * from './components/ui/tooltip';
+export {
+  Tabs as UnderlineTabs,
+  TabsList as UnderlingTabsList,
+  TabsTrigger as UnderlineTabsTrigger,
+  TabsContent as UnderlineTabsContent,
+} from './components/ui/underline-flat-tabs';
+
+export * from './lib/utils';
```
1egoman, line 31: I added a root `src/index.tsx` reexporting everything in `@srcbook/components`. I suspect this is probably not something that anyone would want to use because it makes bundle splitting more more difficult, but on the other hand it seemed like there should be _something_ importable at the package root.

packages/web/src/components/cells/code.tsx
```diff
@@ -270,609 +334,54 @@ export default function CodeCell(props: Props) {
...
+          document
+            .getElementById(file_response.result.filename)
+            ?.scrollIntoView({ behavior: 'smooth' });
+        } else {
+          if (file_response.error) {
+            reject(new Error(`Error fetching file content: ${file_response.result}`));
+          } else {
+            resolve(file_response.result.content);
+          }
+        }
```
1egoman, line 364: I wanted to call out a pretty substantial rework I did of the go to definition here as part of an earlier abstration attempt, but I left it in because it seemed to be a bit cleaner.

In summary, make `onGetDefinitionContents(...)` return a `Promise<string>`, rather than call an ad-hoc callback.

packages/components/src/components/cell-output.tsx
benjreinhart: Sounds good, I believe that is specified in the tsconfig.json `paths` field and also vite.config.ts (though I don't think this package uses vite)

packages/components/src/components/cells/code.tsx
benjreinhart: This is good in that all arguments to this function are now made explicit. Depending on how many levels deep this component tree goes, there might be a lot of noise as a result of the prop drilling. If this becomes tedious, we might be able to:

1. Pass all props explicitly into the root
2. Have the root create a context+hook with all the props
3. Remove some prop passing and allow children to utilize the hook

If that ends up being helpful and we need to reuse this component in different contexts, there may be an even better solution which is to place a dependency on a hook defined elsewhere and at build time swap out that dependency with different implementations for the different contexts.

All that said, this is a longer term thing, we shouldn't try to do this now.

packages/components/src/components/cell-output.tsx
1egoman: It is, yes, but that doesn't seem to convert to relative paths in the built output.

benjreinhart (approved):
All working well for me!

1egoman:
^ Rebased on top of `main` to fix some merge conflicts!

PR #330, opened 2024-10-02
fix CSS cm-editor * + dark:secondary-text + add new compact UI

versecafe:
image

> Some extra CSS changes to add warning and error tooltip backgrounds in followup PR

### DIFF
--- a/packages/web/src/index.css
+++ b/packages/web/src/index.css
@@ -29,7 +29,6 @@
     --sb-yellow-30: 33 97% 85%;
     --sb-yellow-40: 34 96% 81%;
     --sb-yellow-50: 35 95% 76%;
-    --sb-yellow-50-subtle: 35 95% 76% 0.1;
     --sb-yellow-60: 33 76% 69%;
     --sb-yellow-70: 35 31% 45%;
     --sb-yellow-80: 34 28% 32%;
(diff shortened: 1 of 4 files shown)


packages/web/src/index.css
```diff
@@ -37,6 +38,7 @@
     --sb-red-20: 0 100% 86%;
     --sb-red-30: 0 100% 76%;
     --sb-red-40: 0 100% 66%;
+    --sb-red-40-subtle: 0 100% 66% 0.1;
```
benjreinhart, line 41: Why are the existing colors not sufficient here?

versecafe: needed alpha channel variant, got rid of these and moved to `hsl(var(--sb-*-*) / 0.1)`

packages/web/src/index.css
```diff
@@ -390,23 +394,28 @@
...
 }
 
 .cm-tooltip {
   @apply overflow-hidden;
   @apply !bg-background;
   @apply border;
-  @apply rounded-sm;
   @apply shadow-md;
   @apply !border-border;
+  font-size: 14px;
```
nichochar, line 409: isn't 14px the same size as the editor? In which case this is not useful. Didn't we want it to be smaller?

versecafe: it gets override by tw anyways but just pulled it since it's empty junk

nichochar:
Code LGTM, but idk about the screenshot line-height. I'll take over the PR and get this through, since it's getting subjective.

nichochar:
OK, came around after playing around with tighter line-heights, this is actually good as is. Merging!

nichochar (approved):

PR #333, opened 2024-10-06
[FIX]:285 - Type proccess.env based off the available secrets in a no…

BeRecursive22:
…tebook

# Dynamic Environment Type Declarations for Srcbooks

### Initial Creation:

When creating a new srcbook, an empty environment type declaration file (env.d.ts) is generated under /src.
This file initially assumes all environment secrets are disabled.

### Dynamic Updates:

Upon user actions (Add/Enable/Disable/Delete secrets), the system:
a. Retrieves the updated secrets available for the srcbook.
b. Dynamically generates a new env.d.ts file.
c. Writes the updated file to disk under the /src directory.

### Type Safety:

All available environment secrets for a srcbook are typed in the ProcessEnv interface.
This enables TSServer to throw type errors when a user attempts to access non-existent environment secrets.

### Limitations

This feature is not retroactively applicable to srcbooks created prior to this feature release.

### Some assumptions:

- Since env secrets are not created/updated frequently, I chose to write the file instead of reading and appending the types. 
- All the env secrets are disabled initially when a new srcbook is created


BeRecursive22:
@nichochar @benjreinhart I'd like to know your feedback on this approach and whether we can go ahead with this or not.

Type error on hover:

image

Error in console:
image

Suggestions:

image

packages/api/srcbook/index.mts
```diff
@@ -217,3 +228,33 @@ export function removeSrcbook(srcbookDir: string) {
 export function removeCodeCellFromDisk(srcbookDir: string, filename: string) {
   return fs.rm(pathToCodeFile(srcbookDir, filename));
 }
+
+export async function updateSessionEnvTypeDeclarations(sessionId: string) {
+  const sessionSecrets = await getSecretsAssociatedWithSession(sessionId);
+  if (!Object.entries(sessionSecrets).length) return;
+  const envTypeDeclarationFileContent = generateEnvTypesFile(sessionSecrets);
+  const srcbookDir = pathToSrcbook(sessionId);
+  const envDtsPath = Path.join(srcbookDir, 'src', 'env.d.ts');
```
benjreinhart, line 237: Nit but i would put this in the repo root. I think that's where I've typically seen these global envs.

packages/web/vite.config.ts.timestamp-1728221477604-4ec75ba1037b3.mjs
```diff
@@ -0,0 +1,16 @@
+// vite.config.ts
```
benjreinhart, line 1: This file doesn't look like it should be checked in. Can we remove.

BeRecursive22: Sure

benjreinhart (approved):

nichochar:
Neat, thanks @BeRecursive22 !

PR #338, opened 2024-10-08
remove unexpected tsserver messages in server logs

### LINKED ISSUE #291: Receiving unexpected tsserver messages in server logs
I have noticed a lot of tsserver messages in the logs recently that indicate we are still connected to a tsserver instance after we should have shut it down. I'm not exactly sure why this is happening but I didn't use to see these.

```
Received a response for command 'open' and request_seq '0' but no resolver was found. This may be a bug in the code.

Response:
{
  "seq": 0,
  "type": "response",
  "command": "open",
  "request_seq": 0,
  "success": true,
  "performanceData": {
    "updateGraphDurationMs": 379.10445799999997
  }
}

Received a response for command 'open' and request_seq '1' but no resolver was found. This may be a bug in the code.

Response:
{
  "seq": 0,
  "type": "response",
  "command": "open",
  "request_seq": 1,
  "success": true
}

Received a response for command 'open' and request_seq '2' but no resolver was found. This may be a bug in the code.

Response:
{
  "seq": 0,
  "type": "response",
  "command": "open",
  "request_seq": 2,
  "success": true
}
```

swk777:
fix #291 

the cause of the problem was the lack of handling for certain commands that do not require a response, such as open and close.

### DIFF
--- a/.changeset/warm-bugs-approve.md
+++ b/.changeset/warm-bugs-approve.md
@@ -1,0 +1,5 @@
+---
+'@srcbook/api': patch
+---
+
+remove unexpected tsserver messages in server logs
--- a/packages/api/tsserver/tsserver.mts
+++ b/packages/api/tsserver/tsserver.mts
@@ -82,10 +82,10 @@ export class TsServer extends EventEmitter {
   }
 
   private sendWithoutResponse(request: tsserver.protocol.Request) {
-    this.resolvers[request.seq] = () => {}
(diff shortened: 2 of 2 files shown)


BeRecursive22:
@swk777 do we need changeset here or is it okay?

swk777:
> @swk777 do we need changeset here or is it okay?

thx bro

PR #339, opened 2024-10-08
feat: add auto-install dependencies for new typescript srcbook

Sanjoy-droid:
fix: #210 

feat: add auto-install dependencies for new typescript srcbook


nichochar:
I don't think this is the correct fix. Going to close this one in favor of #342 which does it with a config setting.

PR #341, opened 2024-10-08
Update sans font

benjreinhart:
@effyyzhang has chosen a new font for sans: [Instrument Sans](https://fonts.google.com/specimen/Instrument+Sans).


nichochar:
Digging it

PR #342, opened 2024-10-09
Feat/#210 auto install deps when creating a new ts srcbook #210

### LINKED ISSUE #210: Auto-install deps when creating a new TS Srcbook

Aswanth-c:
Fixed [Issue #210](https://github.com/srcbookdev/srcbook/issues/210)

 ### **Changes** 

- Added a toggle button for auto-installing dependencies.
- Added  `AutoInstallation` field in database
- Ran database migrations.
- Added a conditional check and automatic installation of `tsx`, `typescript`, and `@types/node` upon TypeScript Srcbook creation.

 ###  **Question:**

Currently, the packages are installed automatically only upon the creation of a Srcbook. I was not sure if it was a design choice to require the user to press the install button for npm packages, so I have left it as is.  if needed. I can change it to automatically install when the user presses **Run**


packages/api/drizzle/meta/_journal.json
```diff
@@ -78,6 +78,20 @@
       "when": 1726808187994,
       "tag": "0010_create_apps",
       "breakpoints": true
+    },
+    {
+      "idx": 11,
+      "version": "6",
+      "when": 1728457840317,
+      "tag": "0011_apps_external_id_unique",
```
nichochar, line 86: Hmm I wonder if this is a rebase issue, but I expect only one entry in the journal for this PR. Maybe we need to rebase and rerun the migration?

packages/web/src/routes/session.tsx
```diff
@@ -339,6 +343,16 @@ function Session(props: SessionProps) {
...
   ]);
 
+  useEffect(() => {
+    const handleAutoInstallation = () => {
+      if (autoInstallDependencies) {
+        npmInstall();
+      }
+      setInitialInstallAttempt(true);
+    };
+    setTimeout(() => handleAutoInstallation(), 1500);
```
nichochar, line 353: Why is there a timeout set here?

BeRecursive22:
@nichochar @Aswanth-c 

I have observed that if you add cell:format after creating cell:create, the deps are automatically installed. I was wondering why we didn't go with option?

There might be some edge cases...I haven't considered.

nichochar:
Actually, that's the kind of option I was assuming.

Trigger cell:create, and then on callback once things are created, call a `deps:install` event (we shouldn't use cell:format for this as it's semantically a bit weird)

packages/web/src/routes/session.tsx
```diff
@@ -339,6 +343,16 @@ function Session(props: SessionProps) {
     npmInstall,
   ]);
 
+  useEffect(() => {
+    const handleAutoInstallation = () => {
+      if (autoInstallDependencies) {
+        npmInstall();
```
nichochar, line 349: As @BeRecursive22 commented, we have an event we can send up called `deps:install` rather than re-implement the logic here.

This works but for consistency of how we interface with the server, it might be better to dispatch the websocket event. cc @benjreinhart might have an opinion.

packages/api/drizzle/meta/_journal.json
Aswanth-c: On the main branch, the journal entry and snapshot for the 11th migration `0011_apps_external_id_unique.sql` are missing while the migration is present.

packages/web/src/routes/session.tsx
Aswanth-c: I agree that it's a better option to install packages once the code cell is created rather than when the component mounts; I will change that.


` npmInstall();`. function executes `deps:install` under the hood,
isn't it better to call  ` npmInstall();` function?  as it does more checks and error handling

PR #343, opened 2024-10-09
Codeium AI autocomplete integration

1egoman:
Introduces codeium ai autocomplete into the srcbook editor.

Screenshot 2024-10-10 at 3 22 20 PM

A few high level notes:
- There is no way to enable / disable this feature currently.
- The other cells are used as context when performing code generation, leading to more contextually relevant suggestions.
- By default, the `EDITOR_API_KEY` constant token is used. If a user authenticates with codeium, a token linked to their user account is used instead.


packages/web/src/lib/ai-autocomplete/language_server.proto
```diff
@@ -0,0 +1,172 @@
+// NOTE: to generate a new set of json definitions, run:
+// npm run generate-codeium-proto-json
+
+// Copyright Exafunction, Inc.
+
+syntax = "proto3";
+
+package exa.language_server_pb;
+
+import "codeium_common.proto";
```
1egoman, line 10: Here's the underlying protobuf definitions that are used to generate the codeium autocomplete request. These came from here: https://github.com/Exafunction/codeium-react-code-editor/blob/main/exa/language_server_pb/language_server.proto.

Note that to avoid including this raw text in the bundle that is sent to the client, I converted the protobuf file to json descriptiors ([more about these here](https://github.com/protobufjs/protobuf.js?tab=readme-ov-file#using-json-descriptors)). These are a fair bit smaller and work just as well.

Another option I considered but ultimately decided against given the complexity of introducing was an alternate protobuf library that does code generation at build time, like [`buf`](https://buf.build/docs/introduction). If protobufs were going to be used everywhere and definitions were going to be changing often, I think it would be a lot of sense, but given this is only used here that seemed like overkill.

packages/web/src/routes/settings.tsx
```diff
@@ -205,6 +213,80 @@ function Settings() {
...
+                <Button asChild variant="secondary">
+                  <Link
+                    to={`https://www.codeium.com/profile?response_type=token&redirect_uri=${codeiumCallbackUrl}&state=a&scope=openid%20profile%20email&redirect_parameters_type=query`}
+                  >
+                    Authenticate with Codeium
+                  </Link>
+                </Button>
+              </div>
+            )}
+          </div>
```
1egoman, line 289: I added a new interface element to the settings page to facilitate a user authenticating with codeium:

Screenshot 2024-10-10 at 3 07 50 PM

packages/web/package.json
```diff
@@ -10,23 +10,28 @@
     "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
     "format": "prettier --write .",
     "preview": "vite preview",
-    "check-types": "tsc"
+    "check-types": "tsc",
+    "generate-codeium-proto-json": "node --eval 'console.log(\"export default\", JSON.stringify(require(\"protobufjs\").loadSync(\"src/lib/ai-autocomplete/language_server.proto\").toJSON(), null, 2));' > src/lib/ai-autocomplete/languageServerProto.ts"
```
1egoman, line 14: I added this npm task to take the `language_server.proto` file and convert it into the json descriptors data that is required by the client.

Note that currently this generated json descriptors file is committed. I could potentially set this up to generate fresh when starting up the dev server, but as this in practice will rarely / more likely never change, it seemed like setting up more complex build infrastructure here would be less useful than it otherwise would be.

packages/web/src/lib/ai-autocomplete/language_server.proto
versecafe: @1egoman are we okay with the copyright on this?

packages/web/src/lib/ai-autocomplete/codeium_common.proto
```diff
@@ -0,0 +1,166 @@
+// Copyright Exafunction, Inc.
```
benjreinhart, line 1: @1egoman do you know if we're able to pull this in given the copyright?

benjreinhart (approved):
Nit: I would prefer the if the protos we're in their own folder named something like "generated"

packages/web/src/routes/settings.tsx
nichochar: Nit: Could add a link to their sign in page

PR #347, opened 2024-10-11
skeleton stream res hook, stripped down to allow binding stream logic…

versecafe:
… for any completions architechture

### DIFF
--- a/packages/api/server/http.mts
+++ b/packages/api/server/http.mts
@@ -2,7 +2,7 @@ import Path from 'node:path';
 import { posthog } from '../posthog-client.mjs';
 import fs from 'node:fs/promises';
 import { SRCBOOKS_DIR } from '../constants.mjs';
-import express, { type Application, type Response } from 'express';
+import express, { type Application, type Response, type Request } from 'express';
 import cors from 'cors';
 import {
   createSession,
@@ -36,6 +36,8 @@ import { pathToSrcbook } from '../srcbook/path.mjs';
 import { isSrcmdPath } from '../srcmd/paths.mjs';
 import { loadApps, loadApp, createApp, serializeApp, deleteApp } from '../apps/app.mjs';
 import { CreateAppSchema } from '../apps/schemas.mjs';
+import { StreamData, streamText } from 'ai';
+import { createAnthropic } from '@ai-sdk/anthropic';
 
 const app: Application = express();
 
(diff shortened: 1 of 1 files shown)


benjreinhart (commented):
Will defer this pr to @nichochar

packages/api/server/http.mts
```diff
@@ -45,6 +47,37 @@ router.use(express.json());
 
 router.options('/file', cors());
 
+/**
+ * Note things added with data.append will be 2: in stream, LLM res is 0: so you can
```
nichochar, line 51: I don't understand this comment. Can we reword it?

packages/api/server/http.mts
```diff
@@ -45,6 +47,37 @@ router.use(express.json());
...
+ * add any data you need to pass to the client as this goes to trigger UI events or
+ * other actions.
+ */
+router.get('/ai/anthropic/generate', cors(), async (_req: Request, res: Response) => {
```
nichochar, line 55: Why are we hardcoding anthropic here?

nichochar (commented):
Can you give more context on what this is for?

I'm confused why we're hardcoding Anthropic.

I think we should be using vercel's ai library to do streaming?

packages/api/server/http.mts
versecafe: I can swap it, but you mentioned sticking to one model since they handled prompts differently and that 3.5 sonnet was the best so I went with it, nothing stops swapping the name doesn't particularly matter

packages/api/server/http.mts
versecafe: let me think of a good way to phrase it but essentially the streams look like this

```
2:["initialized call"]
0:"To"
0:" store a user ID,"
0:" times of creation and update"
0:", email, and username"
0:", you can use a"
0:" database or"
0:" a data"
0:" structure"
0:" depending"
0:" on your specific needs"
0:". Here's an example"
2:["random interject", "with more data"]
0:" of"
0:" how you might"
0:" structure"
0:" this data in different"
0:" formats"
0:":\n\n1"
0:". SQL"
0:" Database Table"
0:":\n\n```sql"
0:"\nCREATE TABLE users ("
0:"\n    i"
```

in here only 0:"<content here>" will be in the display stream while 2:[...] will just be data sent down to trigger things

versecafe:
> Can you give more context on what this is for?
> 
> I'm confused why we're hardcoding Anthropic.
> 
> I think we should be using vercel's ai library to do streaming?

This does use vercels ai sdk you swap create anthropic for openai or grok and it works the same way? the stream data, append, etc are all agonstic, it's just injecting the correct provider using anthropic for this example

```ts
import { StreamData, streamText } from 'ai';
import { createAnthropic } from '@ai-sdk/anthropic'; // just swap this for @ai-sdk/....
```

packages/api/server/http.mts
```diff
@@ -45,6 +46,11 @@ router.use(express.json());
 
 router.options('/file', cors());
 
+router.options('/examples', cors());
```
nichochar, line 49: This is a typo I think?
Can we call the endpoint `ai/apps/edit` ? There will be another (no streaming) that will be `ai/apps/create` that I'll add on top

PR #349, opened 2024-10-11
Add logic to represent booting state within preview panel and parse stdout from vite to get port

1egoman:
Currently, the preview panel in the app builder is a bit unreliable - when it opens, the iframe sometimes is pointing to the wrong vite server or nothing at all.

This pull request attempts to make this behavior more consistient, by doing three things:
1. When the vite server starts, the srcbook code which invokes the process monitors the stdout coming from the process, and picks up the port from stdout using a regex. This port is then what the client opens, not a hardcoded port like how the logic worked previously.
2. When the panel is opened and the `preview:start` message is sent to the server, the server now sends back a `preview:status` message including a `status` of `"booting"`. - the client then shows a loading indicator. Once a port is detected in the stdout, a secondary message is sent to the client including this port, and the loading indicator goes away, and an iframe is rendered going to this port.
3. When the app builder is reloaded, the port previously extracted from `1.` is sent back to the client again, resulting in the preview panel opening back up to the correct url.

### DIFF
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -145,9 +145,9 @@ export function register(wss: WebSocketServer) {
         JSON.stringify([
           topic,
           'preview:status',
-          existingProcess ? (
-            { status: 'running', url: `http://localhost:${existingProcess.port}/` }
-          ) : { url: null, status: 'stopped' },
+          existingProcess
+            ? { status: 'running', url: `http://localhost:${existingProcess.port}/` }
+            : { url: null, status: 'stopped' },
         ]),
       );
     });
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -18,9 +18,16 @@ import { loadApp } from '../../apps/app.mjs';
 import { fileUpdated, pathToApp } from '../../apps/disk.mjs';
 import { vite } from '../../exec.mjs';
 
+const VITE_PORT_REGEX = /Local:.*http:\/\/localhost:([0-9]{1,4})/;
+
 type AppContextType = MessageContextType<'appId'>;
 
-const processes = new Map<string, ChildProcess>();
+type ProcessMetadata = {
+  process: ChildProcess;
+  port: number | null;
+};
+
+const processMetadata = new Map<string, ProcessMetadata>();
 
 async function previewStart(
   _payload: PreviewStartPayloadType,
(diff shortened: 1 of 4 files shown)


packages/api/server/channels/app.mts
```diff
@@ -18,9 +18,16 @@ import { loadApp } from '../../apps/app.mjs';
...
+
 type AppContextType = MessageContextType<'appId'>;
 
-const processes = new Map<string, ChildProcess>();
+type ProcessMetadata = {
+  process: ChildProcess;
+  port: number | null;
+};
+
+const processMetadata = new Map<string, ProcessMetadata>();
```
1egoman, line 30: Note that now, `processes` is called `processMetadata` and includes a process plus associated metadata (currently just the port it is running on).

packages/web/src/components/apps/workspace/preview.tsx
```diff
@@ -6,15 +6,27 @@ type PropsType = {
...
-  }
+  switch (status) {
+    case "booting":
+    case "connecting":
+      return (
+        <div className={cn("flex justify-center items-center w-full h-full", props.className)}>
+          <span className="text-tertiary-foreground">Booting...</span>
+        </div>
+      );
+    case "running":
```
1egoman, line 19: Right now, this is what the loading state looks like. An official design (I couldn't find one for something like this) would be appreciated!

Screenshot 2024-10-11 at 4 44 14 PM

packages/api/server/channels/app.mts
```diff
@@ -18,9 +18,16 @@ import { loadApp } from '../../apps/app.mjs';
 import { fileUpdated, pathToApp } from '../../apps/disk.mjs';
 import { vite } from '../../exec.mjs';
 
+const VITE_PORT_REGEX = /Local:.*http:\/\/localhost:([0-9]{1,4})/;
```
1egoman, line 21: Here's the regex for extracting the port vite picks from stdout.

benjreinhart (approved):

packages/api/server/channels/app.mts
```diff
@@ -18,9 +18,16 @@ import { loadApp } from '../../apps/app.mjs';
...
+
 type AppContextType = MessageContextType<'appId'>;
 
-const processes = new Map<string, ChildProcess>();
+type ProcessMetadata = {
```
nichochar, line 25: Should this also have the status "running"?

And potentially exit code

packages/api/server/channels/app.mts
```diff
@@ -33,41 +40,56 @@ async function previewStart(
...
+  };
+
   const process = vite({
-    // TODO: Configure port and fail if port in use
     args: [],
     cwd: pathToApp(app.externalId),
     stdout: (data) => {
-      console.log(data.toString('utf8'));
+      const encodedData = data.toString('utf8');
+      console.log(encodedData);
```
nichochar, line 71: Why the 2 lines? Is there a behavior change?

nichochar (commented):
LGTM, can we get this in?

nichochar (approved):

PR #355, opened 2024-10-12
Add folder delete and rename behavior

benjreinhart:
Need to fix folder renaming, but otherwise this is good to go.

### DIFF
--- a/packages/api/server/http.mts
+++ b/packages/api/server/http.mts
@@ -42,6 +42,8 @@ import {
   loadFile,
   createFile,
   createDirectory,
+  renameDirectory,
+  deleteDirectory,
 } from '../apps/disk.mjs';
 import { CreateAppSchema } from '../apps/schemas.mjs';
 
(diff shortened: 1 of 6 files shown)


packages/api/server/http.mts
```diff
@@ -515,6 +517,51 @@ router.post('/apps/:id/directories', cors(), async (req, res) => {
   }
 });
 
+router.options('/apps/:id/directories', cors());
+router.delete('/apps/:id/directories', cors(), async (req, res) => {
+  const { id } = req.params;
+
+  // TODO: validate and ensure path is not absolute
```
nichochar, line 524: I've wanted some path cleaning utils also in other places, this is a good idea. It's particularly useful given that LLMs are giving us paths which i have less confidence in.

packages/web/src/components/apps/use-files.tsx
```diff
@@ -183,6 +209,49 @@ export function FilesProvider({ app, channel, rootDirEntries, children }: Provid
...
+      fileTreeRef.current = deleteNode(fileTreeRef.current, entry.path);
+      forceComponentRerender(); // required
+    },
+    [app.id],
+  );
+
+  const renameFolder = useCallback(
+    async (entry: DirEntryType, name: string) => {
+      const { data: newEntry } = await renameDirectory(app.id, entry.path, name);
+      // const oldFile = filesRef.current[entry.path];
```
nichochar, line 232: Do we need to keep this?

nichochar (approved):
LGTM

packages/web/src/components/apps/use-files.tsx
benjreinhart: This is what I need to fix before merging

PR #356, opened 2024-10-13
Add create file and folder within another folder

benjreinhart:
Unfortunately this is a bit buggy. Getting the focus states working smoothly is not an easy task.

### DIFF
--- a/packages/web/src/components/apps/lib/path.ts
+++ b/packages/web/src/components/apps/lib/path.ts
@@ -16,7 +16,7 @@ export function dirname(path: string): string {
     return '.';
   }
 
-  return parts.pop() || '.';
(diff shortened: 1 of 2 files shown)


nichochar:
Damn the logic is really finicky. Wonder if there are some product simplifications we can do.

benjreinhart:
I definitely agree it's finicky, but it's only the focus portion that's finicky which is currently self contained in a component that's only about 20-30 lines of code, so I'm not terribly concerned at the moment. While I think we could shift around the UX to avoid this problem (e.g., in the extreme case you could use pop up a modal to create a new file), I think this is the UX people are used to for editing experiences.

benjreinhart:
I think we can just live with this for a bit and revisit before launching.

PR #362, opened 2024-10-16
Add original value of files, enabling diffs and reverts

nichochar:


packages/api/ai/plan-parser.mts
```diff
@@ -36,12 +38,13 @@ import Path from 'node:path';
  * </plan>
  */
 
-interface File {
+interface FileAction {
```
benjreinhart, line 41: This is fine for now but I think it would be better to reuse our existing file objects (or something similar to it). Ideally we only need two "file" like types. One for a file entry to be used in the file tree (we don't need to load the content for file tree, so that's why that type is different). Another that contains the contents, probably the `FileType` that exists in shared.

We can wrap those types for AI purposes, which the API might respond with something like:

```json
{
  "data": [
    {
      "description": "...",
      "status": "modified",
      "original": {
        "path": "path/to/file.tsx",
        "name": "file.tsx",
        "source": "...",
        "binary": false
      },
      "modified": {
        "path": "path/to/file.tsx",
        "name": "file.tsx",
        "source": "...",
        "binary": false
      }
    },
    {
      "description": "...",
      "status": "created",
      "original": null,
      "modified": {
        "path": "path/to/file.tsx",
        "name": "file.tsx",
        "source": "...",
        "binary": false
      }
    },
  ]
}
```

Doesn't have to be that exactly but we should try to avoid having N representations of files

nichochar: Agreed. I like the nesting.
I also add basename and dirname in there for convenience because the util function to get them is available in node but not in the browser, so I typically "split it" in the backend at some point and add those fields. Any thoughts on this?

PR #363, opened 2024-10-16
Improve chat component polish. Add diff algorithm. Add DiffType

nichochar:
See video for experience:

https://github.com/user-attachments/assets/9e0bb879-f93d-49ec-81c2-7afd69112b08

I'm starting to realize that we want at least _some_ text from the AI before applying the changes. So I will work on adding that in by changing the prompts and the parser.

I also currently really don't like that the history disappears when we leave the app / reload the page. We clearly need to persist the data structures somewhere. Then we can leverage the little history icon to see past "Diffs".

Types still don't feel perfect @benjreinhart so still iterating. Open to suggestions to improve them.


packages/web/src/components/apps/lib/diff.ts
```diff
@@ -0,0 +1,17 @@
+import * as Diff from 'diff';
```
nichochar, line 1: Considered not using a library but the code is non trivial + this is incredibly battle tested.

Picked this one but there were another couple good candidates, notably diff-match-patch from google

PR #364, opened 2024-10-16
Reintroduce templates + prompt updates for high level changeset description

nichochar:
Previously, the AI was creating the whole app from scratch. 

I modified things so that we have a template with tailwind, vite, no eslint and a simpler tsconfig.

I also updated both prompts for creation and editing, with the new <planDescription> node, which we can present to the user in Markdown form in the ChatBox. We currently use the `Markdown` component but this will likely have styling issues, so it's a stop gap. 

We currently run `npm install` twice, which is not super elegant but actually quite efficient: we run it while the AI is generating and it installs vite, tailwind, etc... then the AI might have added a library, so just in case we run it again.

Note that the optimal fix is to get the AI to give us a list of packages to add, and run it only with those. We can get to that later


benjreinhart (approved):

packages/web/src/components/chat.tsx
```diff
@@ -197,7 +205,8 @@ export function ChatPanel(props: PropsType): React.JSX.Element {
 
     const response = await aiEditApp(props.app.id, query);
 
-    const plan = response.data as Array<PlanItemType>;
+    const plan = response.data as PlanType;
+    setHistory((prevHistory) => [...prevHistory, { type: 'plan', content: plan.description }]);
     if (response.data && Array.isArray(response.data)) {
```
benjreinhart, line 210: you changed response.data to no longer be an array, so this `if` block will never run.

I'm changing all this code, so don't worry about a fix. Just wanted to give a heads up it's probably broken right now.

PR #365, opened 2024-10-16
Remove sidebar auto close and add errors panel

1egoman:
Does a few things:
1. Removes the sidebar auto closing behavior I thought was indicated in the mockups. Now, the sidebar being opened / closed is unrelated to the current tab (code/preview).
2. Updates the loading and error states of the preview panel to be closer in line with the figma designs
3. Add a new "errors" pane to the bottom of the screen, in a status bar. When opened, this errors panel shows a list of errors that have occurred for a user to browse. Clicking a "trash" icon in the upper left sweeps away all errors. If an error occurs and the panel is not open, a small badge is shown next to the errors button letting the user know.
4. Added a new error that shows up when vite crashes / stops unexpectedly (defined as exiting with a status code != 0). Right now this shows a generic message, and when expanded, shows all stdout / stderr logs for the process from when it was started until it crashed. This was built with a "npm install failed" error state in mind as well - this is called out in a few places in the code but the infrastructure to power this hasn't been built out yet.

Screenshot 2024-10-16 at 3 53 19 PM
Screenshot 2024-10-16 at 3 53 27 PM
Screenshot 2024-10-16 at 3 53 38 PM
Screenshot 2024-10-16 at 3 54 02 PM


packages/web/src/components/apps/use-logs.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+export interface LogsContextValue {
+  logs: Array<LogMessage>;
+  clearLogs: () => void;
+  addError: (message: Omit<LogMessage, 'timestamp'>) => void;
+  unreadLogsCount: number;
+
+  open: boolean;
+  togglePane: () => void;
+}
```
1egoman, line 20: I added a new `LogsContext` to allow anywhere in the app to easily have access to add new errors to this new panel. Here's what is in this context!

packages/web/src/components/keyboard-shortcuts-dialog.tsx
```diff
@@ -53,6 +53,8 @@ export default function KeyboardShortcutsDialog({
                     description="format code using Prettier"
                   />
                   <ShortcutRow keys={['alt', 'click']} description="go to definition" />
+                  <h5 className="font-semibold pt-6 pb-2">App Builder</h5>
+                  <ShortcutRow keys={['mod', 'shift', 'y']} description="toggle errors panel" />
```
1egoman, line 57: There's a keyboard shortcut to open the errors panel - I used the closest relevant one I could find that wasn't already bound from vscode.

packages/api/server/channels/app.mts
```diff
@@ -63,12 +63,15 @@ async function previewStart(
...
   const process = vite({
     args: [],
     cwd: pathToApp(app.externalId),
     stdout: (data) => {
       const encodedData = data.toString('utf8');
       console.log(encodedData);
+      bufferedLogs.push(encodedData);
 
       const potentialPortMatch = VITE_PORT_REGEX.exec(encodedData);
       if (potentialPortMatch) {
```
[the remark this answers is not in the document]
nichochar: Good comment to leave in the code actually

packages/shared/src/schemas/websockets.mts
```diff
@@ -167,10 +167,15 @@ export const FileDeletedPayloadSchema = z.object({
...
-export const PreviewStatusPayloadSchema = z.object({
-  url: z.string().nullable(),
-  status: z.enum(['booting', 'running', 'stopped']),
-});
+export const PreviewStatusPayloadSchema = z.union([
+  z.object({ url: z.string().nullable(), status: z.enum(['booting', 'running']) }),
+  z.object({
+    url: z.string().nullable(),
+    status: z.literal('stopped'),
+    ok: z.boolean(),
```
nichochar, line 175: What's the semantic meaning of ok? Not immediately obvious to me, so let's leave a comment

packages/web/src/components/apps/statusbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { useHotkeys } from 'react-hotkeys-hook';
+
+import { Button } from '@srcbook/components/src/components/ui/button';
+import { cn } from '@/lib/utils.ts';
+import { LogMessage, useLogs } from './use-logs';
+
+function getLabelForError(error: LogMessage) {
+  switch (error.type) {
+    case 'vite_error':
+      return 'Error running vite preview server!';
```
nichochar, line 12: Nit: don't need the bangs

packages/web/src/components/apps/statusbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  error: LogMessage;
+};
+
+function CollapsibleErrorMessage({ error }: CollapsibleErrorMessageProps) {
+  const [open, setOpen] = useState(false);
+  return (
+    <>
+      <button
+        className={cn(
+          'flex items-center gap-2 font-mono p-2 first:border-t first:border-t-border focus-visible:outline-none ring-inset focus-visible:ring-1 focus-visible:ring-ring',
```
nichochar, line 28: Just checking that we need `first:border-t-border`. I think that color is automatic, unless maybe there's an override that I'm missing, you can just remove

packages/web/src/components/apps/statusbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const { logs, clearLogs, unreadLogsCount, open, togglePane } = useLogs();
+
+  useHotkeys('mod+shift+y', () => {
+    togglePane();
+  });
+
+  return (
+    <>
+      <div className="flex items-center justify-between h-8 border-t border-t-border px-2">
+        <Button
```
nichochar, line 60: Nit: logic here is a little harder to follow. Would it be more clear to use a:
```
{ open ? (
 .... // open tsx
) : (
 .... // closed tsx
)}

File: packages/web/src/components/apps/use-logs.tsx [hunk 8]
```diff
@@ -0,0 +1,75 @@
...
+    setUnreadLogsCount((n) => n + 1);
+  }
+
+  function togglePane() {
+    setOpen((n) => !n);
+    setUnreadLogsCount(0);
+  }
+
+  // If vite crashes, then create an error log
+  useEffect(() => {
```
nichochar, line 51: Fine for now (I don't have a better suggestion), but somehow feels like a weird place to have this listener

packages/web/src/components/apps/workspace/editor/editor.tsx
```diff
@@ -9,24 +9,31 @@ import { useFiles } from '../../use-files';
...
+  useEffect(() => {
+    if (!openedFile) {
+      return;
+    }
+    setTab('code');
+  }, [openedFile]);
+
   return (
-    <div className="flex flex-col">
+    <div className="grow shrink flex flex-col">
```
nichochar, line 32: Curious: what does `grow shrink` do?

packages/shared/src/schemas/websockets.mts
1egoman: I'll change it to something more descriptive. To answer your question though, the intent was to store whether the process stopped successfully.

packages/web/src/components/apps/statusbar.tsx
1egoman: What you've described is what I am doing in effect a little lower, though instead of no longer rendering a `div`, I'm adjusting its height so that it animates in:

```typescript
<div
  className={cn('flex flex-col overflow-auto cursor-text', 'h-0', { 'h-[320px]': open })}
>
  ...
</div>
```

I'm not sure there's much I can do to simplify the logic you are calling out - this stuff controls whether the "unread" badge is visible and whether the delete icon is visible. As originally conceived, those are a little more nuanced in when they are shown/hidden!

packages/web/src/components/apps/workspace/editor/editor.tsx
1egoman: Makes the container that wraps the editor grow/shrink to take up all space in the flex layout. This became necessary due to the statusbar at the bottom making the layout a bit more complex.

nichochar (approved):

PR #368, opened 2024-10-16
Implement diff modal

benjreinhart:
This also fixes a lot of bugs and warnings with the chat component.

### TODO

- [ ] This does not yet account for files being created or deleted. Need to follow up.

Screenshot 2024-10-16 at 2 18 20 PM

### DIFF
--- a/packages/web/src/components/apps/diff-modal.tsx
+++ b/packages/web/src/components/apps/diff-modal.tsx
@@ -1,0 +1,82 @@
+import { cn } from '@/lib/utils';
+import { Button } from '@srcbook/components';
+import {
+  Dialog,
+  DialogContent,
+  DialogTitle,
+  DialogDescription,
+} from '@srcbook/components/src/components/ui/dialog';
+import { Undo2Icon, XIcon } from 'lucide-react';
+import type { FileDiffType } from './types';
+import { DiffSquares, DiffStats } from './diff-stats';
+import { DiffEditor } from './editor';
+
+type PropsType = {
+  onUndoAll: () => void;
+  onClose: () => void;
+  files: FileDiffType[];
+};
+
+export default function DiffModal({ files, onClose, onUndoAll }: PropsType) {
+  return (
+    <Dialog
+      open
+      onOpenChange={(open) => {
+        if (open === false) {
+          onClose();
+        }
+      }}
+    >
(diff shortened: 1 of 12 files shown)


packages/web/src/components/apps/diff-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+        if (open === false) {
+          onClose();
+        }
+      }}
+    >
+      <DialogContent
+        className={cn('w-[95vw] h-[95vh] max-w-none p-0 gap-0 flex flex-col')}
+        hideClose
+      >
+        {/* Got browser console warnings without this */}
```
nichochar, line 34: I have given up on using the `DialogDescription` and `DialogHeader` and such because of these annoying warnings. They're setup as `<p>` which creates headaches

packages/web/src/components/apps/diff-modal.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </Button>
+      </div>
+    </div>
+  );
+}
+
+function FileDiff({ file }: { file: FileDiffType }) {
+  return (
+    <div className="border rounded-md">
+      <div className="h-10 px-3 flex items-center border-b border-border">
```
nichochar, line 70: I _think_ `border-border` is the default and therefore optional

packages/web/src/components/apps/use-files.tsx
```diff
@@ -23,19 +23,19 @@ import {
 
 export interface FilesContextValue {
   fileTree: DirEntryType;
-  openFile: (entry: FileEntryType) => void;
-  createFile: (dirname: string, basename: string, source?: string) => void;
-  renameFile: (entry: FileEntryType, name: string) => void;
-  deleteFile: (entry: FileEntryType) => void;
-  openFolder: (entry: DirEntryType) => void;
+  openedFile: FileType | null;
```
nichochar, line 26: 🙏

nichochar (approved):
Only nits, approving. Looking good!

packages/web/src/components/apps/diff-modal.tsx
benjreinhart: They emit browser errors/warnings if not there.

PR #376, opened 2024-10-17
Effy update home + global font sizes

effyyzhang:
image
CleanShot 2024-10-16 at 20 22 53@2x

Notebook before and after:
CleanShot 2024-10-17 at 10 44 25@2x
CleanShot 2024-10-16 at 20 28 55@2x


effyyzhang:
CleanShot 2024-10-16 at 20 41 13@2x

packages/components/src/components/ui/dialog.tsx
```diff
@@ -84,7 +84,7 @@ const DialogDescription = React.forwardRef<
 >(({ className, ...props }, ref) => (
   <DialogPrimitive.Description
     ref={ref}
-    className={cn('text-sm text-tertiary-foreground', className)}
+    className={cn('text-sm font text-tertiary-foreground', className)}
```
nichochar, line 87: I dont think this does anything

packages/web/src/index.css
```diff
@@ -154,41 +154,47 @@
 
 @layer components {
   .title {
-    font-size: 3rem;
+    font-size: 2rem;
```
nichochar, line 157: Not super critical given that we want to focus on apps, but do these affect the notebook look?

nichochar (commented):
One `font` I think we should remove, and maybe a screenshot of the old notebook to make sure things don't look broken, but otherwise this LGTM!

packages/web/src/index.css
effyyzhang: This changes the title size of the notebook, no impact on the app. I changed this cause the rest of the font are smaller now.

nichochar (approved):
LGTM

PR #379, opened 2024-10-17
Update file tree entry types to include dirname and basename

benjreinhart:
File tree types now include:

* `path` the full path -- can be used as unique file/folder id (this existed already)
* `dirname`
* `basename` (this existed previously as `name`)

This is part of an effort to make file and file tree types A) very easy to use and B) as similar to one another as possible.

This PR also fixes a bug when inserting a node in the file tree introduced #373 where we removed the ability to insert a new node into the tree.


nichochar:
I think this might have introduced a regression where whenever you apply an AI change (in other words, calling `createFile` triggers a re-render which switches the tab back to the code, instead of the preview.

PR #381, opened 2024-10-17
Adds mechanism to support managed npm install runs from within the app builder

1egoman:
This change adds a new context which can be used from within the app builder experience to programatically run `npm install` or `npm install <packagename>`. This builds on top of much of the existing package.json management logic originally built for srcbook notebooks. Also note that logs stream in while the npm install occurs live!

In addition, I've introduced a very barebones interface that can be used to try it out. Note that a more thought through interface is definitely the long term intention - I just wanted to get something out to get folks moving, and plan to follow up with something later here.

Finally, I've made some small tweaks to the statusbar panel to fix some rendering weirdnesses I encountered.

## High level walkthrough
Screenshot 2024-10-17 at 3 56 26 PM

https://github.com/user-attachments/assets/853f93ea-7d29-4a20-9e6f-1f2a6eb6d7e6

Screenshot 2024-10-17 at 3 57 44 PM

Screenshot 2024-10-17 at 3 58 25 PM

### DIFF
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -169,8 +169,6 @@ async function dependenciesInstall(
       });
     },
   });
-
-  processMetadata.set(app.externalId, { process, port: null });
 }
 
 async function onFileUpdated(payload: FileUpdatedPayloadType, context: AppContextType) {
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -148,7 +148,7 @@ async function dependenciesInstall(
     return;
   }
 
-  const process = npmInstall({
+  npmInstall({
     args: [],
     cwd: pathToApp(app.externalId),
     packages: payload.packages ?? undefined,
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -8,6 +8,8 @@ import {
   FileUpdatedPayloadType,
   PreviewStartPayloadType,
   PreviewStopPayloadType,
+  DependenciesInstallPayloadType,
+  DependenciesInstallPayloadSchema,
 } from '@srcbook/shared';
 
 import WebSocketServer, {
@@ -16,7 +18,7 @@ import WebSocketServer, {
 } from '../ws-client.mjs';
 import { loadApp } from '../../apps/app.mjs';
 import { fileUpdated, pathToApp } from '../../apps/disk.mjs';
-import { vite } from '../../exec.mjs';
+import { vite, npmInstall } from '../../exec.mjs';
 
 const VITE_PORT_REGEX = /Local:.*http:\/\/localhost:([0-9]{1,4})/;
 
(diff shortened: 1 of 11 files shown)


packages/api/server/channels/app.mts
```diff
@@ -135,6 +137,42 @@ async function previewStop(
...
+        log: { type: 'stderr', data: data.toString('utf8') },
+      });
+    },
+    onExit: (code) => {
+      conn.reply(`app:${app.externalId}`, 'dependencies:install:status', {
+        status: code === 0 ? 'complete' : 'failed',
+        code,
+      });
+    },
+  });
```
1egoman, line 171: Here's where the npm install command is executed in the api service.

packages/web/src/components/apps/panels/settings.tsx
```diff
@@ -1,3 +1,37 @@
...
+            Status: <code>{status}</code>
+          </span>
+          <pre className="font-mono text-sm bg-tertiary p-2 overflow-auto rounded-md border">
+            {/* FIXME: disambiguate between stdout and stderr in here using n.type! */}
+            {output.map((n) => n.data).join('\n')}
+          </pre>
+        </>
+      ) : null}
+    </div>
+  );
```
1egoman, line 36: Again just wanted to call out here that this is an interim interface! Clicking the two buttons should allow folks to test both the `npm install` and `npm install <packagename>` paths.

packages/web/src/components/apps/use-package-json.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+type NpmInstallStatus = 'idle' | 'installing' | 'complete' | 'failed';
+
+export interface PackageJsonContextValue {
+  npmInstall: (packages?: string[]) => void;
+  status: NpmInstallStatus;
+  installing: boolean;
+  failed: boolean;
+  output: Array<OutputType>;
+}
```
1egoman, line 18: A lot of this interface should look familiar to the similar context that exists for srcbook notebooks - I originally tried to go down the path of making that reusable, but that proved to be too challenging to do quickly.

packages/web/src/components/apps/workspace/preview.tsx
```diff
@@ -41,11 +41,6 @@ export function Preview(props: PropsType) {
 
       return (
         <div className={cn('w-full h-full', props.className)}>
-          <div className="absolute">
-            <button onClick={() => addError({ type: 'vite_error', contents: 'Bogus error' })}>
-              Trigger
-            </button>
-          </div>
           <iframe className="w-full h-full" src={url} title="App preview" />
         </div>
```
1egoman, line 45: 😬 I forgot I left this in when I merged my [last change](https://github.com/srcbookdev/srcbook/pull/365)... sorry!

1egoman:
Also worth calling out: there's nothing right now serverside to prevent two `npm install`s from happening concurrently. I can probably add some more client side validation to make that case impossible but I don't know how much of a real problem this actually is in practice.

packages/web/src/components/apps/workspace/preview.tsx
nichochar: We're moving fast and I love it. Tiny price to pay

packages/api/server/channels/app.mts
```diff
@@ -150,6 +186,7 @@ export function register(wss: WebSocketServer) {
     .channel('app:<appId>')
     .on('preview:start', PreviewStartPayloadSchema, previewStart)
     .on('preview:stop', PreviewStopPayloadSchema, previewStop)
+    .on('dependencies:install', DependenciesInstallPayloadSchema, dependenciesInstall)
```
nichochar, line 189: Nit: in the notebook we use `deps:install`. I prefer normalizing.

packages/web/src/components/apps/use-package-json.tsx
nichochar: Sounds good. We can take both later and extract an optimal interface, but this is not urgent at all. Good call

nichochar (approved):
LGTM. Some minor nits on naming mostly, but going to merge and fix them myself. Thanks @1egoman

packages/shared/src/schemas/websockets.mts
```diff
@@ -179,3 +179,19 @@ export const PreviewStatusPayloadSchema = z.union([
 
 export const PreviewStartPayloadSchema = z.object({});
 export const PreviewStopPayloadSchema = z.object({});
+
+export const DependenciesInstallPayloadSchema = z.object({
```
nichochar, line 183: This is awefully close to `DepsInstallPayloadSchema`:
```
export const DepsInstallPayloadSchema = z.object({
    packages: z.array(z.string()).optional(),
});
```
might merge them

nichochar: done

nichochar (commented):
Followed up [here](https://github.com/srcbookdev/srcbook/pull/384/files). Only naming. Feature works well!

PR #388, opened 2024-10-18
Ability to update app name

benjreinhart:
This also:

* Introduces app context because that object is A) going to be used in many places and B) we now have some functions to encapsulate that update the object.
* Moves the app header react component out of "editor" because it's not a header just for the editor.

Screenshot 2024-10-18 at 11 34 03 AM


nichochar:
Would prefer one of those "click and it becomes and input which saves on blur", but this is def good first pass

PR #389, opened 2024-10-18
Add an explicit "npm install" loading state, and start to gate features behind if node_modules exists

1egoman:
This pull request does a few things:
1. Implements (mostly) the package installation states in figma.
2. Adds a new `deps:clean` action that allows the app to remove the `node_modules` directory
  a. I introduced this largely for my own testing. If this isn't actually a good idea to have in the app long term, I can remove it.
4. Adds a new `deps:status` and `deps:status:response` action that allows the app to get information about the installed dependencies. Right now this returns an object with a `nodeModulesExists` key, which is used by the app to determine if the user has installed dependencies yet.
5. Makes some adjustments to the preview tab to take into account this new installation state. Namely, instead of auto installing dependencies like before, it now blocks and the user is expected to click the "install" button in the new popup in the lower left corner.

Note that this still doesn't rename `settings` to `packages` or refine the panel's interface - I ran out of time today. I will get to that on monday! There's also some additional interface guarding work left.

https://github.com/user-attachments/assets/609befb0-b448-4a1a-941c-ddd07ed5a4f1

### DIFF
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -1,3 +1,5 @@
+import path from 'node:path';
+import fs from 'node:fs/promises';
 import { ChildProcess } from 'node:child_process';
 
 import {
@@ -10,6 +12,8 @@ import {
   PreviewStopPayloadType,
   DepsInstallPayloadType,
   DepsInstallPayloadSchema,
+  DepsClearPayloadType,
+  DepsStatusPayloadSchema,
 } from '@srcbook/shared';
 
 import WebSocketServer, {
@@ -19,6 +23,7 @@ import WebSocketServer, {
 import { loadApp } from '../../apps/app.mjs';
 import { fileUpdated, pathToApp } from '../../apps/disk.mjs';
 import { vite, npmInstall } from '../../exec.mjs';
+import { directoryExists } from '../../fs-utils.mjs';
 
 const VITE_PORT_REGEX = /Local:.*http:\/\/localhost:([0-9]{1,4})/;
 
@@ -167,7 +172,51 @@ async function dependenciesInstall(
         status: code === 0 ? 'complete' : 'failed',
         code,
       });
+
+      if (code === 0) {
+        conn.reply(`app:${app.externalId}`, 'deps:status:response', {
+          nodeModulesExists: true,
+        });
+      }
     },
+  });
+}
+
+async function clearNodeModules(
+  _payload: DepsClearPayloadType,
+  context: AppContextType,
+  conn: ConnectionContextType,
+) {
+  const app = await loadApp(context.params.appId);
+
+  if (!app) {
+    return;
+  }
+
+  const appPath = pathToApp(app.externalId);
+  const nodeModulesPath = path.join(appPath, 'node_modules');
+  await fs.rm(nodeModulesPath, { recursive: true, force: true });
+
+  conn.reply(`app:${app.externalId}`, 'deps:status:response', {
+    nodeModulesExists: false,
+  });
+}
+
+async function dependenciesStatus(
+  _payload: DepsClearPayloadType,
+  context: AppContextType,
+  conn: ConnectionContextType,
+) {
+  const app = await loadApp(context.params.appId);
+
+  if (!app) {
+    return;
+  }
+
+  const appPath = pathToApp(app.externalId);
+  const nodeModulesPath = path.join(appPath, 'node_modules');
+  conn.reply(`app:${app.externalId}`, 'deps:status:response', {
+    nodeModulesExists: await directoryExists(nodeModulesPath),
   });
 }
 
@@ -187,6 +236,8 @@ export function register(wss: WebSocketServer) {
     .on('preview:start', PreviewStartPayloadSchema, previewStart)
     .on('preview:stop', PreviewStopPayloadSchema, previewStop)
     .on('deps:install', DepsInstallPayloadSchema, dependenciesInstall)
+    .on('deps:clear', DepsInstallPayloadSchema, clearNodeModules)
+    .on('deps:status', DepsStatusPayloadSchema, dependenciesStatus)
     .on('file:updated', FileUpdatedPayloadSchema, onFileUpdated)
     .onJoin(async (topic, ws) => {
       const app = await loadApp(topic.split(':')[1]!);
--- a/packages/web/src/components/apps/header.tsx
+++ b/packages/web/src/components/apps/header.tsx
@@ -60,6 +60,13 @@ export default function EditorHeader(props: PropsType) {
           }}
         />
       )}
+
+      {npmInstallStatus === "installing" ? (
+        <div className="fixed top-0 left-0 right-0 z-[51] h-0.5 pointer-events-none">
+          <div className="h-full w-full bg-white animate-indeterminate" />
+        </div>
+      ) : null}
+
       <header
         className={cn(
           'w-full flex items-center justify-between bg-background z-50 text-sm border-b border-b-border relative',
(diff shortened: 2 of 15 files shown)


packages/web/src/components/apps/header.tsx
```diff
@@ -57,6 +60,13 @@ export default function EditorHeader(props: PropsType) {
           }}
         />
       )}
+
+      {npmInstallStatus === 'installing' ? (
+        <div className="fixed top-0 left-0 right-0 z-[51] h-0.5 pointer-events-none">
+          <div className="h-full w-full bg-white animate-indeterminate" />
+        </div>
+      ) : null}
+
```
1egoman, line 69: I added this indeterminate loading state bar thing across the top. I am kinda a fan of it in this case (a global state of some sort making it very clear to the user something is going on). I tried a smaller "chip" type thing in the header bar as well.

Let me know your thoughts. It would not be hard to pull this out if desired.

packages/web/src/components/apps/use-package-json.tsx
```diff
@@ -1,13 +1,20 @@
...
 
 export interface PackageJsonContextValue {
-  npmInstall: (packages?: string[]) => void;
+  npmInstall: (packages?: string[]) => Promise<void>;
+  clearNodeModules: () => void;
+
+  nodeModulesExists: boolean | null;
   status: NpmInstallStatus;
   installing: boolean;
   failed: boolean;
```
1egoman, line 20: Something I really am not a fan of here: `nodeModulesExists` is `boolean | null` - it is null before it is set by the server initially.

What sucks about this is that any check of this value must explicitly check `=== true`, `=== false`, etc since both `null` and `false` are falsey.

I wanted to get this out to minimize merge conficts. But, I think in hindsight this should have been an enum like `'loading' | 'enabled' | 'disabled'` or something like that. Feel free to make that change, or if not, I'll probablhy make it on monday!

packages/web/src/components/apps/use-package-json.tsx
```diff
@@ -24,41 +31,69 @@ type ProviderPropsType = {
...
+    return () => {
+      channel.off('deps:status:response', callback);
+    };
+  }, [channel]);
+
   const npmInstall = useCallback(
-    (packages?: Array<string>) => {
+    async (packages?: Array<string>) => {
       // NOTE: caching of the log output is required here because socket events that call callback
       // functions in here hold on to old scope values
```
1egoman, line 56: Note that `npmInstall` now returns a promise, so `await npmInstall(); foo` can be used to run something after a `npm install` has finished!

packages/web/src/components/apps/use-preview.tsx
```diff
@@ -43,14 +46,25 @@ export function PreviewProvider({ channel, children }: ProviderPropsType) {
...
 
+  // If the node_modules directory gets deleted, then stop the preview server
+  useEffect(() => {
+    if (nodeModulesExists !== false) {
+      return;
+    }
+    stop();
+  }, [nodeModulesExists]);
+
   // When the page initially loads, start the vite server
```
1egoman, line 68: If the `node_modules` directory gets deleted, the preview server will automatically stop.

packages/web/src/components/apps/workspace/preview.tsx
```diff
@@ -4,27 +4,38 @@ import { useEffect, useState } from 'react';
 import { Loader2Icon } from 'lucide-react';
 import { useLogs } from '../use-logs';
 import { Button } from '@srcbook/components/src/components/ui/button';
+import { usePackageJson } from '../use-package-json';
 
 type PropsType = {
   isActive?: boolean;
+  onShowPackagesPanel: () => void;
   className?: string;
 };
```
1egoman, line 13: I don't beleive this `onShowPackagesPanel` prop is used anymore - I meant to get rid of it. I'll do that on monday, or feel free to do it for me.

packages/web/src/routes/apps.tsx
```diff
@@ -55,20 +56,24 @@ export function AppsPage() {
...
+            <PreviewProvider channel={channelRef.current}>
               <Apps />
-            </PackageJsonProvider>
-          </LogsProvider>
-        </PreviewProvider>
+            </PreviewProvider>
+          </PackageJsonProvider>
+        </LogsProvider>
       </FilesProvider>
     </AppProvider>
```
1egoman, line 67: NOTE: this context ordering is important! `PreviewProvider` uses `PackageJsonProvider`, and `PackageJsonProvider` uses `LogsProvider`.

1egoman:
I also want to call out: I've noticed a lot of instability in the logic that kicks off running `vite`. It seems like sometimes it locks up / becomes unresponsive, and I'm not 100% sure why. I don't _think_ it's related to something I've introduced in this change as it also seems to happen on `main`.

My best lead is that in `packages/api/package.json`, `dev` is set to `vite-node -w ...` - that `-w` causes node to get live reloaded. Maybe srcbook isn't properly killing all of it's subprocesses? More investigation needs to be done here.

packages/web/src/routes/apps.tsx
nichochar: Good type of comment to add in code. Nice to find that if you're about to refactor.

nichochar:
"Packages need to be installed" is a subtle affair, we did some work for that in the original notebook.
This PR introduces the high level "does `node_modules` exist check. Worth noting that we have a library and some utils in there called `depcheck` that you can run and it will tell you if packages need to be installed. It is slightly smarter as it knows to check for the state where some libraries are installed, and `node_modules` exists, but the package.json listed deps are not all there (using package-lock as a source of truth).

I'm bringing this up because we might want different levels of checks. I think checking for node_modules and blocking the server preview is a solid first pass though, and so this is something we can build on.

packages/api/server/channels/app.mts
```diff
@@ -167,10 +172,54 @@ async function dependenciesInstall(
         status: code === 0 ? 'complete' : 'failed',
         code,
       });
+
+      if (code === 0) {
```
nichochar, line 176: In terms of the UI. Does sending this event earlier make a difference with a flash state?

packages/web/src/components/apps/header.tsx
nichochar: Generally like the idea of having a "status" somewhere in the header for installing packages, loading the devserver, or awaiting an AI response. Will see how this feels but in terms of design but props wise 👍

packages/web/src/components/apps/package-install-note.tsx
```diff
@@ -0,0 +1,126 @@
+import { useEffect, useState } from 'react';
+import { CircleAlertIcon, InfoIcon, Loader2Icon, XIcon } from 'lucide-react';
+
+import { usePackageJson } from './use-package-json';
+import { useLogs } from './use-logs';
+import { Button } from '@srcbook/components/src/components/ui/button';
+import { cn } from '@/lib/utils';
+
+const PackageInstallNote: React.FunctionComponent = () => {
```
nichochar, line 9: Nit: in our component terminology, this is a `Toast` (arguably maybe a Card).

packages/web/src/components/apps/use-package-json.tsx
nichochar: One of my good mentors said that nearly all booleans should just be enums :), agree we should switch it. Not super urgent but minor tech debt in its current state

packages/web/tailwind.config.js
```diff
@@ -191,12 +191,17 @@ module.exports = {
           from: { height: 'var(--radix-collapsible-content-height)' },
           to: { height: '0' },
         },
+        indeterminate: {
```
nichochar, line 194: neat trick

nichochar (approved):
Going to merge this and address some of my own comments. Directionally this is good. I'll try to give a clear status by Monday on where things are at.

PR #396, opened 2024-10-21
fix: trash-icon color modified 

### LINKED ISSUE #395: Change the trash icon color when hover to cards under Your Notebooks section
Title: Change Trash Icon Color on Hover in the "Your Notebooks" Section

Description:
Currently, the trash icon in the "Your Notebooks" section does not change color when hovered, which may cause it to be less noticeable as an interactive element. This issue requests a visual enhancement where the trash icon's color changes on hover, providing clearer feedback to the user that it is clickable. The change should only apply to the cards under the "Your Notebooks" section and ensure consistency with the overall design.

Sanjoy-droid:
@benjreinhart 

small ui fix #395 

Along with Trash-Icons in Your Notebooks sections I have also modified in Secrets and Navbar Section.

can checkout in the screednrecord

[srcbook.webm](https://github.com/user-attachments/assets/875625aa-1a0c-485a-9d0c-769bcc77c437)

### DIFF
--- a/packages/web/src/components/cells/generate-ai.tsx
+++ b/packages/web/src/components/cells/generate-ai.tsx
@@ -62,7 +62,7 @@ export default function GenerateAiCell(props: {
           ? 'ring-1 ring-ai-ring border-ai-ring'
           : 'focus-within:ring-1 focus-within:ring-ring focus-within:border-ring',
         error &&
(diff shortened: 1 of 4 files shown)


benjreinhart:
Thanks for this, however we work with a designer and these are not our designs. Additionally, the implementation here does not use our red color. I'm going to close this.

PR #402, opened 2024-10-21
Use routes for code and preview

benjreinhart:
* Default /apps/:id to the preview tab
* Add two routes:
    1. /apps/:id/files (if user has last file opened saved in local storage, redirect to that. Otherwise, show empty state) 
    2. /apps/:id/files/:path (show the file in the editor)

Lots of changes required to untangle things and make them configurable starting at the router.


nichochar (approved):
LGTM

PR #404, opened 2024-10-22
Improve logs and preview server behavior

benjreinhart:
When playing around with the app, I saw some vite dev server behavior I wasn't sure on and then the logs (somewhat lacking) made it more confusing to me. I wasn't sure if we had tried to re-boot it, or what. I also didn't see anything about it shutting down. Then I noticed some websocket events were firing multiple times.

This PR:

1. Removes the duplicated websocket events
2. Fixes a bug where we are replying to a specific connection rather than broadcasting during the lifetime of a running preview server. This has a couple issues:
    1. That connection may no longer exist (e.g., user refreshes page)
    2. This would only forward output to one client. If I have multiple tabs open, I wouldn't see the output in the other tab(s).
3. Adds consistent logging for the preview server states (booting, running, stopping)
4. Cleans the logs. Adds a "srcbook" source to differentiate from vite or other output.

### Initial boot
Screenshot 2024-10-21 at 9 05 14 PM

### Stopped the server which caused it to auto-restart
Screenshot 2024-10-21 at 9 05 43 PM

### DIFF
--- a/packages/api/server/channels/app.mts
+++ b/packages/api/server/channels/app.mts
@@ -39,7 +39,7 @@ const processMetadata = new Map<string, ProcessMetadata>();
 async function previewStart(
   _payload: PreviewStartPayloadType,
   context: AppContextType,
-  conn: ConnectionContextType,
+  wss: WebSocketServer,
 ) {
   const app = await loadApp(context.params.appId);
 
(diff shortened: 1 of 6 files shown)


packages/api/server/channels/app.mts
```diff
@@ -50,39 +50,34 @@ async function previewStart(
   const existingProcess = processMetadata.get(app.externalId);
 
   if (existingProcess) {
-    conn.reply(`app:${app.externalId}`, 'preview:status', {
+    wss.broadcast(`app:${app.externalId}`, 'preview:status', {
```
benjreinhart, line 53: I realize this is sorta confusing, but the `conn` object represents one specific connection and thus `conn.reply` only replies to that one connected client, whereas `wss.broadcast` broadcasts to all clients connected over a given topic.

One of the problems this can cause here is that, since this is being held in a closure for the duration of a long-lived process, the connection may close long before the preview server is done running (refreshing the page will cause this). After the connection is closed, all preview server output would never make its way to any client.

cc @1egoman @nichochar

packages/api/server/channels/app.mts
```diff
@@ -111,11 +105,12 @@ async function previewStart(
     onExit: (code) => {
       processMetadata.delete(app.externalId);
 
-      conn.reply(`app:${app.externalId}`, 'preview:status', {
+      console.log('HERE???', code);
```
benjreinhart, line 108: Whoops, I'll remove this.

packages/web/src/routes/apps/preview.tsx
```diff
@@ -56,11 +56,11 @@ function Preview() {
     case 'stopped':
       return (
         <div className="flex justify-center items-center w-full h-full">
-          {lastStoppedError === null ? (
-            <span className="text-tertiary-foreground">Stopped preview server.</span>
+          {exitCode === null || exitCode === 0 ? (
```
benjreinhart, line 59: I improved this a bit here: https://github.com/srcbookdev/srcbook/commit/e94f15791ea9b6d57f9f174f0c3d3cd07a1dfaea

nichochar:
Thanks for all the touch ups!

PR #405, opened 2024-10-22
Fix undo/redo history affecting multiple files after switching in codeMirror

swk777:
https://github.com/user-attachments/assets/94f065ba-a788-48e1-902a-a8acd73cd7d3


nichochar (approved):
Thanks for the fix

PR #410, opened 2024-10-22
Install npm packages after plan ran. External link option

nichochar:
Ability to open the vite app in a new tab (i want this to show something off in the video).

Also when creating an app, if there were packages they wouldn't be installed and it would crash. I fix that by installing only the right packages post AI plan.

CleanShot 2024-10-22 at 13 28 12


packages/api/ai/plan-parser.mts
```diff
@@ -160,3 +160,13 @@ export async function parsePlan(
...
   }
 }
+
+export function getPackagesToInstall(plan: Plan): string[] {
+  return plan.actions
+    .filter(
+      (action): action is NpmInstallCommand =>
+        action.type === 'command' && action.command === 'npm install',
+    )
+    .map((action) => action.packages)
```
benjreinhart, line 170: .map().flat() can be expressed in one iteration using `flatMap()`.

packages/web/src/components/apps/header.tsx
```diff
@@ -170,25 +171,46 @@ export default function EditorHeader(props: PropsType) {
...
-                  </TooltipTrigger>
-                  <TooltipContent>Stop dev server</TooltipContent>
-                </Tooltip>
-              </TooltipProvider>
+              <>
+                {url && (
+                  <TooltipProvider>
+                    <Tooltip>
+                      <TooltipTrigger asChild>
+                        <Button
```
benjreinhart, line 179: This should just be a link rather than a button with window.open

packages/web/src/components/apps/header.tsx
```diff
@@ -170,25 +171,46 @@ export default function EditorHeader(props: PropsType) {
...
+                          variant="icon"
+                          size="icon"
+                          onClick={() => window.open(url as string, '_blank')}
+                          className="active:translate-y-0"
+                          disabled={previewStatus === 'booting' || previewStatus === 'connecting'}
```
benjreinhart, line 184: nit but probs better to do `previewStatus !== 'running'`, especially since at some point stop may keep the preview stopped (rather than client automatically booting it back up again, which is currently awkward behavior given that we have a stop button that doesn't really stop it).

packages/web/src/components/apps/header.tsx
```diff
@@ -170,25 +171,46 @@ export default function EditorHeader(props: PropsType) {
...
+                    <TooltipTrigger asChild>
+                      <Button
+                        variant="icon"
+                        size="icon"
+                        onClick={() => {
+                          stopPreview();
+                          props.onChangeTab('code');
+                        }}
+                        className="active:translate-y-0"
+                        disabled={previewStatus === 'booting' || previewStatus === 'connecting'}
```
benjreinhart, line 205: same re: previewStatus !== running

PR #411, opened 2024-10-22
Effy small UI polish

effyyzhang:
Adds a new onboarding screen which gates a user from visiting the application's home page until they've added an ai api key:

Screenshot 2024-10-24 at 3 59 06 PM


packages/components/tsconfig.json
```diff
@@ -7,6 +7,6 @@
     "jsx": "react-jsx",
     "baseUrl": "."
   },
-  "include": ["src/index.tsx", "src/**/*.ts", "src/**/*.tsx"],
+  "include": ["src/index.tsx", "src/**/*.ts", "src/**/*.tsx", "../web/src/components/onboarding-modal.tsx"],
```
benjreinhart, line 10: Why this change? FWIW we cannot have files in the `components` package reference files outside it since it does not depend on web (and can't since that's circular dep). Also FWIW, we're going to remove the components package as that was introduced for reasons that no longer exist.

packages/web/src/components/apps/markdown.tsx
```diff
@@ -1,10 +1,11 @@
...
+import { cn } from '@/lib/utils';
 
 export default function Markdown(props: { source: string; className?: string }) {
   return (
-    <div className={cn('sb-prose-chat px-2', props.className)}>
-      <MarkdownReact gfm>{props.source}</MarkdownReact>
-    </div>
+    <div
+      className={cn('sb-prose-chat px-2', props.className)}
+      dangerouslySetInnerHTML={{ __html: marked(props.source) }}
```
benjreinhart, line 8: Why this change? Also, we need gfm.

packages/web/src/routes/home.tsx
```diff
@@ -66,6 +68,29 @@ export default function Home() {
   const [appToDelete, setAppToDelete] = useState<AppType | null>(null);
   const [showCreateAppModal, setShowCreateAppModal] = useState(false);
 
+  const { aiProvider, openaiKey, anthropicKey } = useSettings();
+  const [showOnboarding, setShowOnboarding] = useState(false);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const checkApiKey = () => {
+      const hasSeenOnboarding = localStorage.getItem('hasSeenOnboarding') === 'true';
```
benjreinhart, line 77: This should live in persistent storage (sqlite), at least in the medium-long term.

benjreinhart (commented):
Can we use css for the gradient? The image seems unnecessary and adds to the package sizes.

packages/web/src/routes/home.tsx
1egoman: I updated this logic to instead gate on whether `aiEnabled` being outputted from `useSettings` was enabled rather than segregating the onboarding state individually.

I don't think there's a reason that tracking this separately is important, given that this should be shown for all users that don't have an ai api key set - if so though, please call it out.

benjreinhart (approved):

PR #418, opened 2024-10-23
More robust process handling for apps

benjreinhart:
This PR:

- [x] Stores all app-related processes (npm install, vite dev server) in a more robust map. We can use this to synchronize calls and ensure we only have one process for each running at any given time.
- [x] Fixes the following bug: Preview server doesn't run because no node_modules yet. When this happens, the child process emits an `error` event (containing an error with code 'ENOENT') but no `exit` event. Since it doesn't emit `exit`, we don't call the `onExit` handler and therefore never remove the process from the process map. Subsequent calls to start the server won't work because the process map thinks there's an existing process for the vite server, but that process is dead and never ran. Because it never ran, the port is null and thus we see `locahost:null` as the URL for the dev server.

### TODO

The following still needs to be addressed but will do so in subsequent PR(s)

- [ ] We should either A) remove the [npm install calls in apps.mts](https://github.com/srcbookdev/srcbook/blob/main/packages/api/apps/app.mts#L39-L100) and let the client handle all installing or B) kick off npm install when the app is created on the backend and hook into the process map introduced in this PR ensuring we synchronize npm installs (sketch for this approach is here: https://github.com/srcbookdev/srcbook/pull/412)
- [ ] The client is a bit buggy with npm install calls, sometimes when the npm install toast prompts me, I install and after install finishes (even successfully), i immediately get prompted to install again. I can reproduce this by allowing non AI apps to be created (no prompt in create modal) and also remove the install call in apps.mts.
- [ ] I occasionally still see npm install failing. When this happens, the app gets into a bad state. With this PR, if I re-run npm install on the client, the dev server and other things seem to start working again. But we need to figure out why npm installs fail sometimes

PR #419, opened 2024-10-23
Ensure socket messages are always sent for all npm installs

1egoman:
At a high level, this change should address the first two points in https://github.com/srcbookdev/srcbook/pull/418.

We had discussed previously the notion of whether making all `npm install`s client initiated would be an easy way to fix the first item. However, after getting more into this work, I opted to not do it this way. This was due to a mixture of not being able to come up with an elegant way to pass the state from the creation step to the actual apps interface to invoke the install, and that it seemed like it's likely folks could have multiple instances of a srcbook open and generally getting in the habit of broadcasting events so they all remain in sync no matter which instance caused an install to happen seemed like a good idea.

In more detail, this change:
1. Ensures that for every `npm install` invocation made by the app, all relevant socket events are broadcasted to all connected clients on the given channel. Previously, some calls sent everything, some sent a few events, and some sent none. At least in my anecdotal observations, a lot of complex and hard to reason about situation were coming up due to the server and client's state not being synced.
2. Adds a new message that gets sent when a npm install starts running. This means that all connected app instances will show a "loading" state, even if the npm install in question was initiated on the server or in another app instance.
3. When a user joins a channel initially, send a message to them giving them an initial npm install status, so that if an install is in flight, the newly connected app can display this.

### DIFF
--- a/packages/api/apps/app.mts
+++ b/packages/api/apps/app.mts
@@ -4,11 +4,10 @@ import { type App as DBAppType, apps as appsTable } from '../db/schema.mjs';
 import { applyPlan, createViteApp, deleteViteApp, getFlatFilesForApp } from './disk.mjs';
 import { CreateAppSchemaType, CreateAppWithAiSchemaType } from './schemas.mjs';
 import { asc, desc, eq } from 'drizzle-orm';
-import { deleteAppProcess, npmInstall, waitForProcessToComplete } from './processes.mjs';
+import { npmInstall, waitForProcessToComplete } from './processes.mjs';
 import { generateApp } from '../ai/generate.mjs';
 import { toValidPackageName } from '../apps/utils.mjs';
 import { getPackagesToInstall, parsePlan } from '../ai/plan-parser.mjs';
-import { wss } from '../index.mjs';
 
 function toSecondsSinceEpoch(date: Date): number {
   return Math.floor(date.getTime() / 1000);
@@ -46,20 +45,6 @@ export async function createAppWithAi(data: CreateAppWithAiSchemaType): Promise<
     },
     onExit(code) {
       console.log(`npm install exit code: ${code}`);
-
-      // We must clean up this process so that we can run npm install again
-      deleteAppProcess(app.externalId, 'npm:install');
-
-      wss.broadcast(`app:${app.externalId}`, 'deps:install:status', {
-        status: code === 0 ? 'complete' : 'failed',
-        code,
-      });
-
-      if (code === 0) {
-        wss.broadcast(`app:${app.externalId}`, 'deps:status:response', {
-          nodeModulesExists: true,
-        });
-      }
     },
   });
 
@@ -84,20 +69,6 @@ export async function createAppWithAi(data: CreateAppWithAiSchemaType): Promise<
       },
       onExit(code) {
         console.log(`npm install exit code: ${code}`);
-
-        // We must clean up this process so that we can run npm install again
-        deleteAppProcess(app.externalId, 'npm:install');
-
-        wss.broadcast(`app:${app.externalId}`, 'deps:install:status', {
-          status: code === 0 ? 'complete' : 'failed',
-          code,
-        });
-
-        if (code === 0) {
-          wss.broadcast(`app:${app.externalId}`, 'deps:status:response', {
-            nodeModulesExists: true,
-          });
-        }
       },
     });
   }
@@ -124,9 +95,6 @@ export async function createApp(data: CreateAppSchemaType): Promise<DBAppType> {
     },
     onExit(code) {
       console.log(`npm install exit code: ${code}`);
-
-      // We must clean up this process so that we can run npm install again
-      deleteAppProcess(app.externalId, 'npm:install');
     },
   });
 
--- a/packages/api/apps/app.mts
+++ b/packages/api/apps/app.mts
@@ -8,6 +8,7 @@ import { deleteAppProcess, npmInstall, waitForProcessToComplete } from './proces
 import { generateApp } from '../ai/generate.mjs';
 import { toValidPackageName } from '../apps/utils.mjs';
 import { getPackagesToInstall, parsePlan } from '../ai/plan-parser.mjs';
+import { wss } from '../index.mjs';
 
 function toSecondsSinceEpoch(date: Date): number {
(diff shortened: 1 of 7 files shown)


packages/api/apps/app.mts
```diff
@@ -1,10 +1,10 @@
...
 import { db } from '../db/index.mjs';
 import { type App as DBAppType, apps as appsTable } from '../db/schema.mjs';
-import { applyPlan, createViteApp, deleteViteApp, pathToApp, getFlatFilesForApp } from './disk.mjs';
+import { applyPlan, createViteApp, deleteViteApp, getFlatFilesForApp } from './disk.mjs';
 import { CreateAppSchemaType, CreateAppWithAiSchemaType } from './schemas.mjs';
 import { asc, desc, eq } from 'drizzle-orm';
-import { npmInstall } from '../exec.mjs';
+import { npmInstall, waitForProcessToComplete } from './processes.mjs';
 import { generateApp } from '../ai/generate.mjs';
 import { toValidPackageName } from '../apps/utils.mjs';
```
1egoman, line 9: It seems like this `npmInstall` was missed during the migration previously - so I converted this to use the same common process-managed interface as everything else. In theory, this meant that two `npm install` commands could still run in parallel.

packages/api/apps/app.mts
```diff
@@ -57,9 +56,10 @@ export async function createAppWithAi(data: CreateAppWithAiSchemaType): Promise<
   const packagesToInstall = getPackagesToInstall(plan);
 
   if (packagesToInstall.length > 0) {
+    await waitForProcessToComplete(firstNpmInstallProcess);
+
     console.log('installing packages', packagesToInstall);
-    npmInstall({
-      cwd: pathToApp(app.externalId),
+    npmInstall(app.externalId, {
```
1egoman, line 62: I introduced this `waitForProcessToComplete` function here to ensure that these npm installs during an ai app creation always happen sequentially. In practice, almost always this function should be a noop.

packages/api/apps/processes.mts
```diff
@@ -62,23 +63,84 @@ export function deleteAppProcess(appId: string, process: ProcessType) {
...
+            nodeModulesExists: true,
+          });
+        }
+      },
+    }),
+  };
+  processes.set(appId, newlyStartedProcess);
+
+  return newlyStartedProcess;
 }
```
1egoman, line 144: This might overload this process abstraction a bit, but I moved all the socket message emitting logic here - this is what makes sure these always get emitted for all cases where `npm install` is run.

I could potentially make another level of abstraction here if desired, though I don't think in practice there would be a reason one would want to call `npmInstall` without the socket logic associated.

packages/api/server/channels/app.mts
```diff
@@ -266,5 +240,15 @@ export function register(wss: WebSocketServer) {
...
+    .onJoin((topic, conn) => {
+      // FIXME: there's almost certainly a better way to get the app id than this?
+      const appExternalId = topic.replace(/^app:/, '');
+
+      // When connecting, send back info about an in flight npm install if one exists
+      const npmInstallProcess = getAppProcess(appExternalId, "npm:install");
+      if (npmInstallProcess) {
+        conn.reply(`app:${appExternalId}`, 'deps:install:status', { status: 'installing' });
+      }
+    });
```
1egoman, line 253: When a client joins the channel, send a message letting the client know if there is a npm install in progress.

packages/api/server/ws-client.mts
```diff
@@ -214,7 +214,7 @@ export default class WebSocketServer {
     if (event === 'subscribe') {
       conn.subscriptions.push(topic);
       conn.reply(topic, 'subscribed', { id: payload.id });
-      channel.onJoinCallback(topic, conn.socket);
+      channel.onJoinCallback(topic, conn);
```
benjreinhart, line 217: The special `onJoinCallback` was implemented as a quick hack before #332. However, now that #332 is in place, we should make the callback signature the same as all other handlers:

```diff
+type HandlerType = (
+  payload: Record<string, any>,
+  context: MessageContextType,
+  conn: ConnectionContextType,
+) => void;
+
 /**
  * Channel is responsible for dispatching incoming messages for a given topic.
  *
@@ -52,15 +58,11 @@ export class Channel {
     string,
     {
       schema: z.ZodTypeAny;
-      handler: (
-        payload: Record<string, any>,
-        context: MessageContextType,
-        conn: ConnectionContextType,
-      ) => void;
+      handler: HandlerType;
     }
   > = {};

-  onJoinCallback: (topic: string, ws: WebSocket) => void = () => {};
+  onJoinCallback: HandlerType = () => {};

   constructor(topic: string) {
     this.topic = topic;
@@ -130,7 +132,7 @@ export class Channel {
     return this;
   }

-  onJoin(callback: (topic: string, ws: WebSocket) => void) {
+  onJoin(callback: HandlerType) {
     this.onJoinCallback = callback;
     return this;
   }
@@ -214,7 +216,11 @@ export default class WebSocketServer {
     if (event === 'subscribe') {
       conn.subscriptions.push(topic);
       conn.reply(topic, 'subscribed', { id: payload.id });
-      channel.onJoinCallback(topic, conn.socket);
+      channel.onJoinCallback(
+        payload,
+        { topic: match.topic, event: event, params: match.params },
+        conn,
+      );
       return;
     }
```

The code in server/channels/app.mts can then do:

```typescript
.onJoin((_payload, context, conn) => {
  const appExternalId = (context as AppContextType).params.appId;
  // ...
});
```

packages/api/apps/app.mts
benjreinhart: Instead of `waitForProcessToComplete`, a nicer API might be to have a wrapper around `npmInstall` that returns a promise so that callers like these can just await the install call.

benjreinhart (approved):
Going to approve and we can merge this now but I think we should follow up and remove these backend npm install calls in favor of the client. There may be good reasons to kick off npm installs from BE code in the future, but right now I can't think of those reasons and it seems simpler to implement the following flow:

create app -> immediate redirect to client -> client assumes 'normal' flow, checks for deps, installs if not there, etc.

packages/api/apps/app.mts
nichochar: +1 we never want two concurrent ones, so that guard can be abstracted away

PR #430, opened 2024-10-29
Stream LLM response when editing apps

benjreinhart:
Follow ups:

* Change creation flow immediately redirect to created app and use edit flow to get streaming support
* Stream partial plan description markdown text
* Break response into more granular pieces we can display on client for more responsiveness


nichochar:
LGTM. Given how this works, we should be able to populate the DiffBox temporarily with "updating ${filename}..."

Although we need to be careful to not make that clickable otherwise if users click on the file we'll interrupt the request

nichochar (approved):

nichochar:
I realized that we could also push the file streams another way:
- we could be pushing them through the websocket "file:updated"
- this would mean they update real time
- at the end of the request, we could still send the total diff plan like we did before
I think this might work well, but might present issues based on the ordering of the LLM's response tags

benjreinhart:
Yes for sure, we can improve this a lot beyond this first pass.

PR #432, opened 2024-10-31
Reintroduce streaming. New streaming parser

nichochar:
Fixed up a few bugs.
This thing is very stateful and even though it didn't break during some basic testing, I would not be surprised to find more bugs.

The bugs often have to do with the difficulty of parsing CDATA vs non CDATA sections, how we manage the currentTag stack, and generally mutability of the buffer.

I have ideas on how we can improve it when we reimplement it later.

This PR also removes the "create a version" mechanism in the frontend when there is only an npm install command.


nichochar:
closing, dupe of #433

Issue #442, opened 2024-11-16
Windows users are not able to setup the code locally properly.


aakash-a-dev:
The issue was while setting up the code locally.

![Error](https://github.com/user-attachments/assets/496b1851-1dbd-4d4f-bde1-193fd0217f43)

danielmeloalencar:
remove -rf

PR #443, opened 2024-11-16
For setting up in the local environment in Windows

aakash-a-dev:
Fix: #442 
### **Context**
While working on the `package/shared` directory, running the `pnpm run build` command was functioning correctly in **WSL (Windows Subsystem for Linux)** but failing in **native Windows environments**. This issue occurred because the `rm -rf ./dist` command used in the `prebuild` script is a Unix/Linux command and is not supported natively in Windows.

### **Error Details**
When running `pnpm run build` in the `package/shared` directory on Windows, the following error was encountered:

Error

This prevented the `build` process from completing successfully.

---

## **Resolution**
To address this, the `prebuild` script in `package.json` was updated to use a **cross-platform solution**. This ensures compatibility with both Linux and Windows environments.

### **Changes Made**
1. Replaced the `rm -rf ./dist` command in the `prebuild` script with the `rimraf` package, a cross-platform tool for deleting files and directories.
2. Updated the `package.json` file in the `package/shared` directory.

---

## **Outcome**
- The updated `prebuild` script ensures compatibility across both **Linux** and **Windows** platforms.
- The issue caused by the unsupported `rm -rf ./dist` command is resolved.
- Files now run locally on Windows without errors, and the workflow remains unaffected for Linux users.

---

image

---

This could help to fix the error for upcoming contributors independent of their OS


packages/web/vite.config.ts.timestamp-1731777964119-c1b6e3663557b.mjs
```diff
@@ -0,0 +1,35 @@
+// vite.config.ts
```
nichochar, line 1: What is this file?

nichochar (commented):
adding rimraf LGTM, from what I can read online it's a good solution to the cross platform rm problem. 

I am not sure what that vite.config.ts file is however

aakash-a-dev:
I added worked on this on best of my knowledge but if you have any better approach for this I will be glad to learn and work on that. 
PS: I confirmed that this file is automatically created and may be this is happening in Windows. Adding few resources to check
https://github.com/vitejs/vite/issues/13267
https://stackoverflow.com/questions/76510164/can-i-safely-delete-vite-config-ts-timestamp-files

nichochar (approved):

PR #444, opened 2024-11-18
Fixed: For making TS notebook in Windows

### LINKED ISSUE #297: Creating ts srcbook on windows
When I boot up srcbook and try to click the button to create a typescript srcbook I get the following error on Windows.

Screenshot 2024-09-14 080432

This error led me to the `node_modules/@srcbook/api/dist/tsserver/tsserver.mjs` file. I noticed there was a call to spawn on line 37. If I add the option `shell: true,` to the spawn command then I don't get the error anymore.
```js
const child = spawn('npx', ['tsserver'], {
  cwd:options.cwd,
  shell: true // added to make spawn call work
});
```
Now I'm able to create a typescript srcbook without generating an error. However when I start to type into the text area to create my typescript file I get another error. This error causes the local server to stop running.

Screenshot 2024-09-14 080828

I think this is an issue with the `TextEncoder / TextDecoder` in `node_modules/@srcbook/api/dist/tsserver/messages.mjs`. I tried editing the carriage return char code thinking it was an issue with line endings in Windows.
```js
// const CARRIAGE_RETURN_CHAR_CODE = '\r'.charCodeAt(0);
const CARRIAGE_RETURN_CHAR_CODE = '\r\n'.charcCodeAt();
```

I should note that I'm able to create javascript notebooks and run them without an issue.

aakash-a-dev:
#### **Issue Overview**

The error `spawn npm ENOENT` occurred when trying to spawn a `tsserver` instance using `npx` in a Node.js script. This error typically happens when `npm` is either missing or not found in the system's `PATH`, which is required for `npx` to function correctly. Although `npx` is supposed to work independently, it relies on `npm` being available in the environment. Since the project is using `pnpm` as the package manager, the solution is to use `pnpm dlx` instead of `npx`.

---

### **Before Resolving the Issue the Error**
Error

### **After Resolving the Issue the Error**
Output-2

image

### **How I Resolved the Issue**

1. **Replace `npx` with `pnpm dlx`**:
   
   The command that spawns the `tsserver` instance was originally using `npx`:
   ```ts
   const child = spawn('npx', ['tsserver'], {
     cwd: options.cwd,
     shell: true
   });
   ```
   This was changed to use `pnpm dlx` to make it compatible with `pnpm`:
   ```ts
   const child = spawn('pnpm', ['dlx', 'tsserver'], {
     cwd: options.cwd,
     shell: true
   });
   ```
   The `pnpm dlx` command is similar to `npx` but uses `pnpm` to run the specified package without requiring global installation. It ensures that `pnpm` is handling the execution, avoiding the `spawn npm ENOENT` error.

---

### **Changes Made to `package.json`**

No changes were directly made to `package.json` to resolve the error. However, to ensure that `pnpm dlx` can be used smoothly, make sure the following:

1. **Added `rimraf` Dependency**:
  
   ```json
   "devDependencies": {
     "rimraf": "^6.0.1",
     ...
   }
   ```

---

### **Conclusion**

By switching from `npx` to `pnpm dlx`, the `spawn npm ENOENT` error was resolved. This solution is optimal for projects using `pnpm` as the package manager, ensuring compatibility and proper execution of the `tsserver` without relying on `npm`'s presence in the environment.

PS: This solution doesn't turn off the server, which was mentioned in the issue, and it also gives the output in my Windows system.

Fix: #297


packages/api/tsserver/tsservers.mts
```diff
@@ -44,8 +44,9 @@ export class TsServers {
     // created, the dependencies are not installed and thus this will
     // shut down immediately. Make sure that we handle this case after
     // package.json has finished installing its deps.
-    const child = spawn('npx', ['tsserver'], {
+    const child = spawn('pnpm', ['dlx' ,'tsserver'], {
```
nichochar, line 47: We don't want to do this, because `pnpm` is a dependency for dev, but not for users. This would require our users to have pnpm installed.

aakash-a-dev: Ooo okay, I didn't think that way. I will look into this further. Thank you!

aakash-a-dev:
This approach ensures portability and compatibility by directly resolving the `tsserver` binary from the project's `node_modules`, avoiding reliance on global tools like `pnpm` or `npx`. 

### **Key Features of the Solution**

1. **No Global Dependencies**:  
   - Avoids relying on tools like `pnpm` or `npx`, which require global installation. Only depends on the local `node_modules` directory.

2. **Cross-Platform Compatibility**:  
   - Automatically resolves to `tsserver.cmd` on Windows and `tsserver` on Linux/macOS, ensuring it works across all operating systems.

3. **Version Consistency**:  
   - Always uses the `typescript` version installed in the project, ensuring consistency between development environments and deployments.

4. **Shell-Free Execution**:  
   - Directly runs the `tsserver` binary without relying on shell commands, improving security and reducing potential issues. 

**Output**

image

image

I hope this solution will be helpful for dev and end user both!

Issue #445, opened 2024-11-20
Add support for Google Gemini


glassBead-tc:
Access requested on the Discord. Templating the work after the Grok support item: https://github.com/srcbookdev/srcbook/pull/438

Assign this one to me @nichochar

aakash-a-dev:
I am on it!

PR #446, opened 2024-11-21
Added Gemini Ai Support

aakash-a-dev:
You can check Feature from here

image

Please let me know 

fix: #445

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,11 +20,12 @@
     "@changesets/cli": "^2.27.8",
     "@srcbook/configs": "workspace:^",
     "@types/node": "^20.14.2",
+    "eslint": "^8.57.0",
     "prettier": "^3.3.3",
-    "typescript": "5.6.2",
-    "eslint": "^8.57.0"
+    "typescript": "5.6.2"
   },
   "dependencies": {
+    "@ai-sdk/google": "^1.0.1",
     "turbo": "^2.1.1"
   },
   "packageManager": "pnpm@9.12.1",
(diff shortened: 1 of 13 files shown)


aakash-a-dev:
image
image

Gemini Support Added! I am doing the lining issue for now

aakash-a-dev:
Resolved all the issue, the gemini api is working smoothly. Please do let me know in any case there could be any scope of improvement.

package.json
```diff
@@ -20,9 +20,9 @@
     "@changesets/cli": "^2.27.8",
     "@srcbook/configs": "workspace:^",
     "@types/node": "^20.14.2",
+    "eslint": "^8.57.0",
```
nichochar, line 23: Not a big deal but why change this order?

nichochar (commented):
This all looks good to me! Will test it this weekend before merging but I think this is good to go!

package.json
aakash-a-dev: Ahh, we can ignore this during the integration I was getting a type error which I thought could be resolved by having ai-sdk for Google installed in the root package when I later realized it has nothing to with it, I removed it from package.json for avoiding not so necessary installation in the package.

aakash-a-dev: And honestly I didn't realized that it changed the order.

aakash-a-dev: We don't need this particular commit as well btw! I don't know how it got committed because I remembered I removed the dependency, if it's possible to not align with the following package.json then it works because we have nothing to do with root's package.json file

nichochar (approved):
LGTM, merging!

Issue #451, opened 2024-11-29
How to configure host address (0.0.0.0) for server in Docker environment?


PastaGringo:
### How to configure host address for server in Docker environment?

**Description**
I'm trying to run srcbook in a Docker container, but I can't access it from outside the container. The server seems to listen only on localhost by default, even with proper port mapping in Docker.

**Current Setup**
- Running srcbook in a Docker container
- Using `npx srcbook@latest start`
- Port mapped correctly in docker-compose
- Server starts but only accessible inside container

**Question**
Is there a way to configure the server to listen on 0.0.0.0 instead of localhost? I couldn't find this option in the documentation.

**Environment**
- srcbook version: 0.0.15
- Running in Docker container

glassBead-tc:
Reproduced the issue. Working on a fix now.

glassBead-tc:
PR: https://github.com/srcbookdev/srcbook/pull/460

nichochar:
This is released in 0.0.16. Closing

PR #459, opened 2024-12-04
Fix issue #451: Add support for exposure outside Docker container to other devices on network

glassBead-tc:
Added network exposure configuration to enable accessing Srcbook from other devices when running in Docker. This includes:

- Added docker-compose.yml with 0.0.0.0 binding configuration
- Configured port mapping to expose port 2150 to the network
- Set HOST environment variable for external network access

This change allows users to access their Srcbook instance from other devices on their network when running in Docker.

### DIFF
--- a/.changeset/dirty-hotels-press.md
+++ b/.changeset/dirty-hotels-press.md
@@ -1,0 +1,5 @@
+---
+'srcbook': patch
+---
+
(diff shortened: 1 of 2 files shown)


glassBead-tc:
Updating this to make exposure to other devices on local network optional.

PR #466, opened 2024-12-26
Working for Windows

aakash-a-dev:
Fix: #297
Changes are made but requires to test on different OS.
Proof of work video here:
https://youtu.be/DC6lnsA_itU

### DIFF
--- a/packages/api/exec.mts
+++ b/packages/api/exec.mts
@@ -1,3 +1,4 @@
+/* eslint-disable turbo/no-undeclared-env-vars */
 import Path from 'node:path';
 import { spawn } from 'node:child_process';
 
@@ -5,6 +6,9 @@ interface NodeError extends Error {
   code?: string;
 }
 
+/**
+ * Base type for execution requests containing common properties
+ */
 export type BaseExecRequestType = {
   cwd: string;
   stdout: (data: Buffer) => void;
@@ -13,20 +17,33 @@ export type BaseExecRequestType = {
   onError?: (err: NodeError) => void;
 };
 
+/**
+ * Type for Node.js execution requests
+ */
 export type NodeRequestType = BaseExecRequestType & {
   env: NodeJS.ProcessEnv;
   entry: string;
 };
 
+/**
+ * Type for NPM installation requests
+ */
 export type NPMInstallRequestType = BaseExecRequestType & {
   packages?: Array<string>;
   args?: Array<string>;
 };
 
+/**
+ * Type for NPX execution requests
+ */
 type NpxRequestType = BaseExecRequestType & {
   args: Array<string>;
+  env?: NodeJS.ProcessEnv;
 };
 
+/**
+ * Type for spawn call requests
+ */
 type SpawnCallRequestType = {
   cwd: string;
   env: NodeJS.ProcessEnv;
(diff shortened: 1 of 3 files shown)


packages/api/exec.mts
```diff
@@ -38,9 +55,56 @@ type SpawnCallRequestType = {
...
+      Path.join(process.env.LOCALAPPDATA || '', 'npm'),
+    ];
+
+    const existingPath = finalEnv.PATH || '';
+    const newPaths = additionalPaths.filter((path) => !existingPath.includes(path)).join(';');
+
+    finalEnv.PATH = newPaths + (existingPath ? ';' + existingPath : '');
+  }
+
+  console.log(`Executing command: ${finalCommand}`);
```
nichochar, line 98: Let's remove the debug logs

packages/api/exec.mts
```diff
@@ -1,10 +1,14 @@
+/* eslint-disable turbo/no-undeclared-env-vars */
 import Path from 'node:path';
 import { spawn } from 'node:child_process';
 
 interface NodeError extends Error {
   code?: string;
 }
 
+/**
```
nichochar, line 9: Please remove these comments

aakash-a-dev: Okay on it

aakash-a-dev:
Removed Debugging console & Comments made from my side

aakash-a-dev:
The updated code improves maintainability, efficiency, and cross-platform compatibility by centralizing executable resolution into a reusable `ExecutableResolver` singleton. It leverages `where` on Windows and `which` on Unix to dynamically resolve paths, caching results to avoid redundant calls. This ensures consistent handling of executables like `node`, `tsx`, and `vite` while prioritizing system executables (e.g., those in `Program Files`). Platform-specific differences, such as handling `.cmd` files on Windows, are abstracted, reducing complexity in individual functions. This approach is cleaner, more scalable, and minimizes duplication, making the code easier to extend for additional commands or environments in the future.

packages/api/tsserver/tsservers.mts
```diff
@@ -44,8 +44,14 @@ export class TsServers {
     // created, the dependencies are not installed and thus this will
     // shut down immediately. Make sure that we handle this case after
     // package.json has finished installing its deps.
-    const child = spawn('npx', ['tsserver'], {
+
+    const npxCmd = process.platform === 'win32' ? 'npx.cmd' : 'npx';
+
+    console.log(`Spawning tsserver with ${npxCmd} in ${options.cwd}`);
```
nichochar, line 50: Remove a debug log

packages/api/exec.mts
```diff
@@ -135,38 +157,34 @@ export function tsx(options: NodeRequestType) {
...
-    env: process.env,
+export const vite = ({ cwd, args = [], env = process.env, ...rest }: NpxRequestType) =>
+  spawnCall({
+    command:
+      process.platform === 'win32' ? 'npx.cmd' : Path.join(cwd, 'node_modules', '.bin', 'vite'),
+    cwd,
+    args: process.platform === 'win32' ? ['vite', ...args] : args,
+    env: {
+      ...env,
+      FORCE_COLOR: '1',
```
nichochar, line 187: What's this?

nichochar, line 187: What's this?

packages/api/exec.mts
```diff
@@ -135,38 +157,34 @@ export function tsx(options: NodeRequestType) {
...
-  return spawnCall({
-    ...options,
-    command: Path.join(options.cwd, 'node_modules', '.bin', 'vite'),
-    env: process.env,
+export const vite = ({ cwd, args = [], env = process.env, ...rest }: NpxRequestType) =>
+  spawnCall({
+    command:
+      process.platform === 'win32' ? 'npx.cmd' : Path.join(cwd, 'node_modules', '.bin', 'vite'),
+    cwd,
+    args: process.platform === 'win32' ? ['vite', ...args] : args,
```
nichochar, line 184: So on windows, we need to call the command and then add `vite` as an arg? Feels like this is `vite vite`. Surprising

aakash-a-dev: Have to hard code it, only then it's running otherwise vite is not getting started, atleast for me it havn't

packages/api/exec.mts
aakash-a-dev: ..env is for copying all the object, for creating in the following object although I added FORCE_COLOR: '1' Because it was getting difficult to find the response so for making it shine I added, I will remove it if you want

packages/api/exec.mts
aakash-a-dev: Alien, Yes! While Vite code which was already there didn't started vite and thrown error, require to hard code this in the following way

packages/api/exec.mts
aakash-a-dev: For just making sure I can track that this log can easily noticeable while I was debugging, if you want I will remove FORCE_COLOR it was for me to make sure that it is logging

packages/api/exec.mts
aakash-a-dev: Removed

packages/api/exec.mts
```diff
@@ -38,25 +40,56 @@ type SpawnCallRequestType = {
   onError?: (err: NodeError) => void;
 };
 
-export function spawnCall(options: SpawnCallRequestType) {
-  const { cwd, env, command, args, stdout, stderr, onExit, onError } = options;
-  const child = spawn(command, args, { cwd: cwd, env: env });
+class ExecutableResolver {
```
nichochar, line 43: I'm a bit scared of this change. It's a pretty big refactor of an important part of the app.

I'm wondering if we can remove the new class implementation, and instead have some simple function based routing.

So first run a function `spawnCall` which will call:
`spawnCallUnix` and `spawnCallWindows` based on the process.platform switch you do line 56. 

That way the mac / unix code can stay the exact same which I think is safeer, and windows can have it's own code path.

aakash-a-dev: Let me change it then!

PR #509, opened 2026-07-29
fix: stop malformed input and lost sessions from crashing the server

nichochar:
```
packages/api/
  server/
    ws-client.mts         +38/-3     guard envelope parse, catch handler rejections
  tsserver/
    tsserver.mts          +58/-15    request timeouts, reject on exit, guard stdout parse
  processes.mts           +12/-1     only delete the key you still own; add has()
  test/
    ws-client.test.mts    +232/-0    16 tests — this file had zero coverage
    processes.test.mts    +120/-0    8 tests incl. the double-exec regression
tasks/0003-...md          +24/-2     close
tasks/0005-...md          +15/-1     partial close, record what's left open
.changeset/               +18/-0
```

Stacked on #508. Closes task 0003, partially closes task 0005.

Last PR in the stack. Several inputs could take the whole server down, which matters more than each one's individual severity — the server holds notebook state that hasn't been flushed.

## The shape worth naming

Three of these are the same mistake: **a throw inside an event-emitter callback is an uncaught exception, not a failed operation.** The surrounding `try`/`catch` or promise looks like it covers the code, and doesn't.

| Where | Trigger |
|---|---|
| `handleIncomingMessage`, in a socket `'message'` listener | one malformed websocket frame |
| tsserver's `parse()`, in a `'data'` listener | any framing it doesn't recognise |
| the depcheck callback, in an `exec` callback | fixed earlier in the stack (#508), same shape |

The websocket case is the sharpest: per-event payloads were already `safeParse`d, but the *envelope* was `JSON.parse`d and `zod.parse`d inline. So the careful validation was one line below the unguarded one.

## Also fixed

**Unhandled handler rejections.** Websocket handlers are async and throw freely — `findSession` throws for an unknown id, which a reconnecting client with a stale session triggers by itself. They were invoked unawaited, so that was an unhandled rejection, and Node exits on those by default.

**Process registry clobbering.** Running a cell twice before the first exits replaced the map entry, and then the *first* process's exit handler deleted the key now owned by the *second* — leaving a process running that `cell:stop` insisted didn't exist. The handler now only clears the key if it still holds the same process reference.

**tsserver promises that never settled.** A resolver was stored by `seq` with no timeout and no reject path, so if tsserver died or never answered, hover/completions/go-to-definition hung forever and the resolver map leaked. Now a 10s timeout plus rejection on process exit. 10s because these are interactive — an answer that late is no more useful than none; what matters is that they settle at all.

## Tests

24 new. `ws-client.mts` had **no coverage at all** despite being the dispatch path for every notebook operation, so this adds topic matching and wildcard extraction, subscribe/broadcast/unsubscribe, and each crash case. Suite is 62, up from 11 at the start of the stack.

The handler-error tests silence `console.error` — they deliberately exercise the logging path, and a passing run shouldn't print stack traces that look like failures.

## Left open on purpose

**What double-exec should do.** The registry no longer corrupts itself, but the first process is orphaned rather than killed or refused. That's a product question — replace, refuse, queue? — and it's coupled to the execution-model decision in task 0006, since a stateful kernel would change the answer entirely.

**Whether stop actually kills TypeScript cells.** `SIGTERM` goes to the `tsx` wrapper, not the process group. If `tsx` forks a child node process, stop is only cosmetic for TypeScript cells. Needs a running notebook and a long-running cell to check; noted in task 0005 rather than quietly closed.

**A rejected tsserver request still leaves the client hanging**, because responses are broadcasts with no request correlation. Task 0009.

No top-level `unhandledRejection` handler: with the handler-level catch there's no known path to one, and a blanket handler would hide the next one rather than surface it.

### DIFF
--- a/.changeset/curly-pandas-tap.md
+++ b/.changeset/curly-pandas-tap.md
@@ -1,0 +1,18 @@
+---
+'@srcbook/api': patch
+---
+
+Stop several inputs from crashing the server.
+
+A malformed websocket frame, a websocket message referring to a session that no longer exists,
+or unexpected tsserver output could each end the process. In every case the throw happened
+inside an event-emitter callback, where the surrounding `try`/`catch` could not reach it. They
+are now logged and the operation is dropped.
+
+Running the same cell twice before the first run exited corrupted the process registry: the
+first process's exit handler deleted the entry belonging to the second, leaving a process
+running that stop reported as nonexistent.
+
+tsserver requests (hover, completions, go-to-definition) now time out after 10 seconds and
+reject if the tsserver process exits. They previously waited forever and leaked their
+resolvers.
--- a/packages/api/processes.mts
+++ b/packages/api/processes.mts
@@ -17,8 +17,19 @@ export class Processes {
     this.processes[key] = process;
 
     process.on('exit', () => {
-      delete this.processes[key];
+      // Only forget this process if it is still the one registered under the key.
+      //
+      // Running the same cell twice before the first exits replaces the entry, and
+      // without this check the first process's exit would delete the second's key —
+      // leaving a running process that `kill` then claims doesn't exist.
+      if (this.processes[key] === process) {
+        delete this.processes[key];
+      }
     });
+  }
+
+  has(sessionId: string, cellId: string) {
+    return this.toKey(sessionId, cellId) in this.processes;
   }
 
   kill(sessionId: string, cellId: string) {
--- a/packages/api/server/ws-client.mts
+++ b/packages/api/server/ws-client.mts
@@ -212,8 +212,21 @@ export default class WebSocketServer {
   }
 
   private handleIncomingMessage(conn: ConnectionType, message: RawData) {
-    const parsed = JSON.parse(message.toString('utf8'));
-    const [topic, event, payload] = WebSocketMessageSchema.parse(parsed);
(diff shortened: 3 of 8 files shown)


nichochar:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/srcbookdev/srcbook/509?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#509** <a href="https://app.graphite.com/github/pr/srcbookdev/srcbook/509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/srcbookdev/srcbook/509?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#508** <a href="https://app.graphite.com/github/pr/srcbookdev/srcbook/508?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#507** <a href="https://app.graphite.com/github/pr/srcbookdev/srcbook/507?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

